PR #1, opened 2016-10-03
Awesome IPLD endeavour

diasdavid:


package.json
```diff
@@ -1,8 +1,8 @@
 {
-  "name": "ipfs-merkle-dag",
```
dignifiedquire: don't forget to deprecate the latest versions of these modules on npm with a message that they were renamed

package.json
```diff
@@ -1,8 +1,8 @@
...
-  "version": "0.7.3",
-  "description": "A JavaScript implementations of the IPFS MerkleDAG implementations (protobufs)",
-  "main": "lib/index.js",
+  "name": "ipld-dag-pb",
+  "version": "0.0.1",
+  "description": "JavaScript Implementation of the MerkleDAG Node in Protobuf.",
+  "main": "src/index.js",
```
dignifiedquire: TODO: change before merge

src/resolver.js
```diff
@@ -0,0 +1,92 @@
...
+  } else if (split[0] === 'data') {
+    return { value: node.data, remainderPath: '' }
+  } else {
+    throw new Error('path not available')
+  }
+}
+
+/*
+ * tree: returns a flattened array with paths: values of the project. options
+ * are option (i.e. nestness)
```
dignifiedquire: > options are option

this doesn't sound right

src/resolver.js
```diff
@@ -0,0 +1,92 @@
...
+    })
+  })
+
+  if (node.data && node.data.length > 0) {
+    paths.push({ path: 'data', value: node.data })
+  }
+  return paths
+}
+
+// TODO recheck this API
```
dignifiedquire: what's the plan here?

PR #3, opened 2016-10-26
Async Crypto Endeavour

dignifiedquire:
Ref https://github.com/ipfs/js-ipfs/pull/485

### DIFF
--- a/src/dag-node.js
+++ b/src/dag-node.js
@@ -2,11 +2,13 @@
 
 const stable = require('stable')
 const mh = require('multihashes')
+const waterfall = require('async/waterfall')
+
 const DAGLink = require('./dag-link')
 
 class DAGNode {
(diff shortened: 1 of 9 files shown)


src/dag-node.js
```diff
@@ -137,18 +139,26 @@ class DAGNode {
...
+  multihash (type, callback) {
+    if (typeof type === 'function') {
+      callback = type
+      type = 'sha2-256'
+    }
+    if (!this._cached[type] || this._updated) {
+      waterfall([
+        (cb) => util.serialize(this, cb),
+        (serialized, cb) => util.hash(type, serialized, cb)
+      ], (err, digest) => {
```
diasdavid: Calling it a `digest`  can be misleading, because it is a multihash (it is prefixed with the multicodec)

PR #6, opened 2016-11-14
feat: update DAGLink and DAGNode to have an immutable API

diasdavid:
> **WIP**

@dignifiedquire @VictorBjelkholm @haadcode did a push on this so that we can get this final API stable, while the API for IPLD Formats are still hot, I'm afraid that if we give it a lot more time, people will start using all of that async ugliness because of the lack of immutability here and then it becomes harder to change.

I'm finishing this up, but API is there, would love your review ASAP

- [x] Agree with the API
- [x] Update remaining tests
- [x] Update README

**note:** This PR also cuts the umbilical chord to the merkleDAG version in Go, which created a funky api in JS


src/dag-node.js
```diff
@@ -2,102 +2,134 @@
...
-        return callback(err)
-      }
-      link.name = name
-      this.addRawLink(link)
-      callback()
+    Object.defineProperty(this, 'data', {
+      get () {
+        return data
+      },
+      set: immutableError
```
dignifiedquire: You can make this nicer as you are already using `class`

```js
class Hello {

  get myImmutableProp () {
    return this._secretProp
  }

  set myImmutableProp () {
    throw new Error('fail')
  }
}
```

src/dag-link.js
```diff
@@ -3,28 +3,28 @@
...
+      this.multihash = multihash
     }
-  }
 
-  toJSON () {
-    return {
-      Name: this.name,
-      Size: this.size,
-      Hash: mh.toB58String(this.hash)
+    this.json = {
```
dignifiedquire: This should be defined as immutable as the props in DagNode

src/dag-node.js
diasdavid: well, that is what I did, but defined the function above so that I could reuse it

haadcode:
I have quite a few comments for this, but can't review it atm. Will get back to this asap.

src/dag-node.js
diasdavid: right, nevermind, I got what you meant now :)

src/dag-link.js
diasdavid: agreed 👍

src/dag-node.js
```diff
@@ -2,102 +2,134 @@
...
+            if (typeof multihash === 'string') {
+              multihash = mh.fromB58String(multihash)
+            }
+            // TODO
+            // 1. create link from name + multihash
+            // 2. add to the set of links
+            // 3. create new node
+          }
+        }
+        callback(new Error('invalid arguments'))
```
diasdavid: *

diasdavid:
All right, this is ready to be reviewed so that I can go forward and bubble up updates across the other repos.

Big wins from this PR:

- Got a tons of feedback from @haadcode 
- The whole dagPB became more functional
- Code structure improved, smaller files with clear goals
- Better test coverage
- Documentation was also improved
- Code overall improved a ton

@haadcode @dignifiedquire @VictorBjelkholm would like a final 👍 so that I can move forward.

Thank you!

src/dag-node/index.js
```diff
@@ -0,0 +1,84 @@
...
+  get multihash () {
+    return this._multihash
+  }
+
+  set multihash (multihash) {
+    throw new Error("Can't set property: 'multihash' is immutable")
+  }
+}
+
+exports = module.exports = DAGNode
```
haadcode: This will make `const d = new DAGNode()` possible for users. I thought we didn't want that to be exposed directly so that users are forced to user .create and have no confusion as to how to create DAGNodes.

src/index.js
```diff
@@ -1,7 +1,11 @@
 'use strict'
 
-exports.DAGNode = require('./dag-node.js')
-exports.DAGLink = require('./dag-link.js')
-exports.resolver = require('./resolver.js')
-exports.util = require('./util.js')
+exports.DAGNode = require('./dag-node')
+exports.DAGLink = require('./dag-link')
```
haadcode: Same as above, this will expose DAGLink to the consumers.

src/util.js
```diff
@@ -55,29 +25,42 @@ exports.serialize = (dagNode, callback) => {
...
   })
 
-  stable.inplace(links, exports.linkSort)
-
   const buf = pbn.Data || new Buffer(0)
 
-  const dagNode = new DAGNode(buf, links)
-
-  callback(null, dagNode)
+  DAGNode.create(buf, links, callback)
```
haadcode: @diasdavid we talked about the IPLD format API and this is pretty much what I meant how it forces sub-optimal code: deserialize depends on DAGNode and DAGNode depends on serialize (serialize and deserialize being both in the same file) essentially creating a cyclic dependency between the two. For the sake of getting this shipped, let's not focus on that but keep this in mind for future discussions.

haadcode:
Couple of small comments, otherwise LGTM!

src/util.js
diasdavid: We could break util into separate files for each function, avoiding that 'cyclic' dependency, do note that in CJS, modules loaded are cached, solving this issue without any intervention.

Nevertheless, this is not a IPLD Format issue, if an issue at all, is just because we want the DAGNode instances to be immutable and easy to work with and so, we need the serialize version upon its creation.

src/dag-node/util.js
```diff
@@ -0,0 +1,36 @@
...
+function cloneData (dagNode) {
+  let data = new Buffer(0)
+
+  if (dagNode.data && dagNode.data.length > 0) {
+    data = new Buffer(dagNode.data.length)
+    dagNode.data.copy(data)
+  }
+
+  return data
+}
```
dignifiedquire: this can be simplified to 

```js
function cloneData (dagNode) {
  assert(Buffer.isBuffer(dagNode.data), 'Not a valid dagnode')
  
  const data = new Buffer(dagNode.data.length)
  dagNode.data.copy(data)
  return data
}
```

src/dag-node/util.js
```diff
@@ -0,0 +1,36 @@
...
+
+function cloneLinks (dagNode) {
+  return dagNode.links.length > 0 ? dagNode.links.slice() : []
```
dignifiedquire: no need for the length check, `[].slice() === []`

src/dag-node/util.js
```diff
@@ -0,0 +1,36 @@
...
+}
+
+function linkSort (a, b) {
+  return (new Buffer(a.name || '', 'ascii').compare(new Buffer(b.name || '', 'ascii')))
```
dignifiedquire: This is super wasteful, the link names should be converted to buffers once and then sorted, not on  every sorting call.

src/util.js
```diff
@@ -55,29 +25,42 @@ exports.serialize = (dagNode, callback) => {
...
+  } else {
+    pbn.Links = null
+  }
+
+  return pbn
 }
 
+exports.serialize = serialize
+exports.deserialize = deserialize
+exports.cid = cid
```
dignifiedquire: 👎 on the whole `util` interface, it seems arbitrary and not very useful. If these methods are needed why are they not exposed at the top level?

diasdavid: @dignifiedquire go through https://github.com/ipld/interface-ipld-format and https://github.com/ipld/js-ipld-resolver/pull/60#issuecomment-254094608

dignifiedquire: okay :/

src/dag-link/index.js
```diff
@@ -0,0 +1,61 @@
+'use strict'
+
+const mh = require('multihashes')
+const assert = require('assert')
+
+// Link represents an IPFS Merkle DAG Link between Nodes.
+class DAGLink {
+  constructor (name, size, multihash) {
```
diasdavid: names are not mandatory (see files in unixfs, a file has N chunks, but no names.

src/dag-node/create.js
```diff
@@ -0,0 +1,59 @@
...
+    data = undefined
+  }
+  if (typeof dagLinks === 'function') {
+    callback = dagLinks
+    dagLinks = []
+  }
+  if (typeof hashAlg === 'function') {
+    callback = hashAlg
+    hashAlg = undefined
+  }
```
diasdavid: I would have to create if conditions by number anyway

src/dag-link/index.js
dignifiedquire: But that is not handled properly as `new DAGLink(123, 'myhash')` throws at the moment

PR #7, opened 2016-11-14
Refactor feat/immutable PR

haadcode:
Refactored https://github.com/ipld/js-ipld-dag-pb/pull/6


haadcode:
This PR doesn't pass tests yet, it's WIP but wanted to start the discussion.

src/dag-link.js
```diff
@@ -2,28 +2,57 @@
...
 
-    this.json = {
+  toJSON() {
+    return {
       name: this.name,
       size: this.size,
-      hash: this.multihash ? mh.toB58String(this.multihash) : undefined
+      hash: this.multihash ? mh.toB58String(this.multihash) : null
     }
   }
```
diasdavid: 👍

haadcode:
There's also couple of places where the code needs to handle callback vs. Promises, mainly the DAGNodeFactory methods like create(). Will update/fix that soon.

src/dag-link.js
```diff
@@ -2,28 +2,57 @@
...
+
+  set size() {
+    throw ImmutableError
+  }
+
+  get hash() {
+    return this._multihash ? mh.toB58String(this._multihash) : undefined
+  }
+
+  set hash() {
```
diasdavid: s/hash/multihash

src/dag-link.js
```diff
@@ -2,28 +2,57 @@
...
+
+  get size() {
+    return this._size
+  }
+
+  set size() {
+    throw ImmutableError
+  }
+
+  get hash() {
```
diasdavid: s/hash/multihash

haadcode:
Let me know if you want to hear the reasoning behind the changes.

src/dag-node-factory.js
```diff
@@ -0,0 +1,139 @@
+'use strict'
+
+const stable = require('stable')
+const sort = stable.inplace
+const mh = require('multihashes')
+const util = require('./util')
+const DAGLink = require('./dag-link')
+
+class DAGNodeFactory {
+  static create (data, dagLinks, hashAlg) {
```
diasdavid: Also, I can see the value of having the create function in a separate file, having to do: `dagPB.dagNodeFactory.create(..)`, seems a tad long, just to create an instance, we can scope the `create` function withing dagPB.DAGNode even if it is in a separate file.

src/dag-node-factory.js
```diff
@@ -0,0 +1,139 @@
...
+    if (!hashAlg) {
+      hashAlg = 'sha2-256'
+    }
+
+    const links = dagLinks.map((l) => {
+      if (!l.constructor && l.constructor.name !== 'DAGLink') {
+        return l
+      } 
+
+      // haadcode: are the .name vs .Name for backwards compatibility?
```
diasdavid: yep (go-ipfs outputs json with that)

src/dag-node-factory.js
```diff
@@ -0,0 +1,139 @@
...
+      const link = new DAGLink(l.name || l.Name,
+                               l.size || l.Size,
+                               l.hash || l.Hash || l.multihash)
+      return link
+    })
+
+    // haadcode: would be better to have: links = sort(links, util.linkSort)
+    sort(links, util.linkSort)
```
diasdavid: the sort is a sort inplace, which works well with a immutable property

src/dag-node-factory.js
```diff
@@ -0,0 +1,139 @@
...
+
+      /* 
+        Alternatively, if util functions were promisified, 
+        we could write this much shorter as:
+        
+        return util.serialize({ data: data, links: links})
+          .then((serialized) => util.hash(hashAlg, serialized))
+          .then((multihash) => new DAGNode(data, links, serialized, multihash))
+      */
+    })
```
diasdavid: I don't believe we are going to change this internal component to a Promised based one now. As we discussed in the past and again at the last gathering, so far the guideline is `callbacks at the internals, promises+callbacks at the top level`. 

This will only change if there is a huge discussion and decision here: https://github.com/ipfs/js-ipfs/issues/557

src/dag-node-factory.js
```diff
@@ -0,0 +1,139 @@
...
+
+    if (link) {
+      links.push(link)
+      sort(links, util.linkSort)      
+    } else {
+      throw new Error("Link given as the argument is invalid")
+    }
+
+    // haadcode: shouldn't we clone the serialized and multihash properties too?
+    return DAGNodeFactory.create(data, links, dagNode.serialized, dagNode.multihash)
```
diasdavid: .create should not take the serialized and multihash, that is generated inside create (or it was in the main PR), note that if you add a link, then the serialized and multihash is different, this is probably why this PR is failing the tests.

src/dag-node-factory.js
```diff
@@ -0,0 +1,139 @@
...
+    if (typeof nameOrMultihash === 'string') {
+      links = dagNode.links.filter((link) => link.name !== nameOrMultihash)
+    } else if (Buffer.isBuffer(nameOrMultihash)) {
+      links = dagNode.links.filter((link) => !link.hash.equals(nameOrMultihash))
+    } else {
+      throw new Error('second arg needs to be a name or multihash')
+    }
+
+    // haadcode: shouldn't we clone the serialized and multihash properties too?
+    return DAGNodeFactory.create(data, links, dagNode.serialized, dagNode.multihash)
```
diasdavid: same as https://github.com/ipld/js-ipld-dag-pb/pull/7/files#r87823980

src/dag-node-factory.js
```diff
@@ -0,0 +1,139 @@
...
+  // was: makeLink(node, callback)
+  static toDAGLink (dagNode) {
+    return new DAGLink(null, dagNode.size, dagNode.multihash)
+  }
+
+  static clone (dagNode) {
+    const data = DAGNodeFactory._cloneData(dagNode)
+    const links = DAGNodeFactory._cloneLinks(dagNode)
+    // haadcode: shouldn't we clone the serialized and multihash properties too?
+    return DAGNodeFactory.create(data, links, dagNode.serialized, dagNode.multihash)
```
diasdavid: same as https://github.com/ipld/js-ipld-dag-pb/pull/7/files#r87823980

src/dag-node.js
```diff
@@ -1,281 +1,83 @@
...
 
-function immutableError () {
-  throw new Error('Immutable property')
-}
+const ImmutableError = new Error('Immutable property')
 
 class DAGNode {
-  constructor (data, links, serialized, size, multihash, json) {
+  constructor (data, links, serialized, multihash) {
+    // Should these be "throw new Error()" instead?
```
diasdavid: that is what assert do if it is missing

src/index.js
```diff
@@ -1,6 +1,6 @@
 'use strict'
 
-exports.DAGNode = require('./dag-node.js')
+exports.DAGNode = require('./dag-node-factory.js')
```
diasdavid: we need to expose the class too.

haadcode:
Will address the feedback, push and come back to the comments

diasdavid:
Thank you for giving a thorough review and coming up with a new structure. I've made a fair amount of comments, it would be ideal if tests were passing so that we know the PR is correct (you will see comments showing on why it's not).

Things I like and will certainly adopt for https://github.com/ipld/js-ipld-dag-pb/pull/6:

- Avoiding Object.defineProperty, as also mentioned by @dignifiedquire (https://github.com/ipld/js-ipld-dag-pb/pull/6#discussion_r87779504)
- Factoring out the create function, so that the DAGNode file is less long
- Make .json a .toJSON sync function

Things I'm not in favour and given our previous discussions, they won't be included in the main PR

- Migration to Promises for internals as noted and justified at : https://github.com/ipld/js-ipld-dag-pb/pull/7#discussion_r87823617
  - Note: It might make sense to support also a promises API for the .create function since it is a top level consumed object (from ipfs.object.get), so that people don't have their promise chain broken, however, we are not dropping callbacks.

Things that don't make sense to me to have throughout our codebase:

- Define "Classes with static methods just to export a bunch of functions" instead of exporting functions with the conventional module.exports. It is mostly a style/convention change which doesn't add any benefit.

haadcode:
The PR now passes tests. Linter is happy except one (too many nested callbacks in tests), which was there from the original PR.

src/dag-node.js
```diff
@@ -1,281 +1,83 @@
...
-
-    if (this.data && this.data.length > 0) {
-      dataClone = new Buffer(this.data.length)
-      this.data.copy(dataClone)
+  toJSON() {
+    return {
+      data: this.data,
+      links: this.links.map((l) => l.json),
+      hash: mh.toB58String(this.multihash),
+      size: this.size
```
haadcode: Note that the size comes from `this.size`:
```javascript
get size () {
  return this.links.reduce((sum, l) => sum + l.size, this.serialized.length)
}
```

src/index.js
haadcode: We can't expose the DAGNode class. If we do that, DAGNode will become *mutable*, ie. doing `const node = new DAGNode(...)` becomes possible which is not what we want.

The general idea of the factory pattern is to expose only methods for creating the DAGNodes, not for users to directly use DAGNode as it was before. This is the first step to immutability.

haadcode:
> Migration to Promises for internals as noted and justified at : #7 (comment)
Note: It actualy might make sense to support also a promises API for the .create function since it is a top level consumed object (from ipfs.object.get), so that people don't have their promise chain broken, however, we are not dropping callbacks.

I removed the Promises from all the methods. I do think we should expose them with Promises, too, as we do in all our public facing APIs. I assume users will need to create DAGNodes themselves (ie. not used *just* internally), right?

> Define "Classes with static methods just to export a bunch of functions" instead of exporting functions with the conventional module.exports. It is mostly a style/convention change which doesn't add any benefit.

This is matter of a) preference and b) code style. If we use classes, we should use them consistently. If we use functions and prototypical inheritance, then use that consistently across the project. As for preference, I personally think wrapping the functions to static methods in a class is much more clearer than exporting functions (looks cleaner and makes the namespacing intuitive).

haadcode:
I also updated the README to reflect how the DAGNode would be used with the changes in this PR but in short:
```javascript
dagPB.DAGNode.create(..., cb)
```

It's the same as in the original PR.

src/dag-node.js
diasdavid: Since it is immutable, we don't need to do this ops everytime, just on creation

src/index.js
diasdavid: We can force the users to always use the 'create', but exposing the Class does not make any instance mutable though.

diasdavid:
Rad! We are converging! Made a couple of comments :)

src/dag-node.js
haadcode: You're right. Same for the other toJSON() props. Will change it so that they all get calculated in the constructor. A small speed improvement is a big speed improvement :)

src/index.js
haadcode: If we expose the DAGNode class, it'll be possible for users to use it, ie. use the constructor. The only way to force not to do that is to not expose it and that, like said, is the first step to an immutable data structure. This is a generally known and widely-used pattern: don't expose the data structure (for use) but rather provide an interface that *returns* the data structure (DAGNode in our case).

Is there a reason why we would *need* to expose it?

haadcode:
@diasdavid what do you think about changing dag-node-factory.js to dag-node-utils.js and DAGNodeFactory to DAGNodeUtils? I originally named it Fatory thinking it'll only create DAGNodes but since we have all the utils (addLink, etc.) in there, I think DAGNodeUtils would make more sense.

src/dag-node-factory.js
```diff
@@ -0,0 +1,193 @@
...
+const stable = require('stable')
+const sort = stable.inplace
+// const mh = require('multihashes')
+const protobuf = require('protocol-buffers')
+const proto = protobuf(require('./dag.proto'))
+const util = require('./util')
+const DAGNode = require('./dag-node')
+const DAGLink = require('./dag-link')
+
+class DAGNodeFactory {
```
diasdavid: Let's move this to 
```
module.exports = create
function create ...
```

src/dag-node-factory.js
```diff
@@ -0,0 +1,193 @@
...
+  static create (data, dagLinks, hashAlg, callback) {
```
diasdavid: wrap the create in `promisify`, so that it supports both.
plus add a test for promises

src/dag-node-factory.js
```diff
@@ -0,0 +1,193 @@
...
+      callback = data
+      data = undefined
+    }
+    if (typeof dagLinks === 'function') {
+      // empty obj
+      callback = dagLinks
+      dagLinks = []
+    }
+    if (typeof hashAlg === 'function') {
+      // empty obj
```
diasdavid: what does this comment mean?

src/dag-node-factory.js
```diff
@@ -0,0 +1,193 @@
...
+
+class DAGNodeFactory {
+  static create (data, dagLinks, hashAlg, callback) {
+    if (typeof data === 'function') {
+      // empty obj
+      callback = data
+      data = undefined
+    }
+    if (typeof dagLinks === 'function') {
+      // empty obj
```
diasdavid: what does this comment mean?

src/dag-node-factory.js
```diff
@@ -0,0 +1,193 @@
...
+
+    if (newLink) {
+      links.push(newLink)
+      sort(links, util.linkSort)
+    } else {
+      throw new Error('Link given as the argument is invalid')
+    }
+
+    DAGNodeFactory.create(data, links, callback)
+  }
```
diasdavid: Not sure if we want this methods in a thing that is called factory.

Not only forces people to have to require the module if they want to add a link:

```
const node = <some dagnode I got from ipfs.object.get
// now I want to add a link
const dagNodeFactory = require('ipld-dag-pb').dagNodeFactory
dagNodeFactory.addLink(node...)
```

while it could just be (as it was):

```
const newNode = node.addLink(link)
```

src/dag-node-factory.js
```diff
@@ -0,0 +1,193 @@
...
+
+  static removeLink (dagNode, nameOrMultihash, callback) {
```
diasdavid: same as `addLink`

src/dag-node.js
```diff
@@ -1,281 +1,79 @@
...
-      }
+    this._data = data || new Buffer(0)
+    this._links = links || []
+    this._serialized = serialized || new Buffer(0) // TODO: default serialized object
+    this._multihash = multihash || new Buffer(0) // TODO: default multihash object
+    this._json = {
+      data: this.data,
+      links: this.links.map((l) => l.json),
+      hash: mh.toB58String(this.multihash),
+      size: this.size
```
diasdavid: This makes it so that folks can change the values by doing `node._data = something` while in the previous case that would not happen. Are we happy with that?

src/dag-node.js
```diff
@@ -1,281 +1,79 @@
...
-        )
-      }
-    })
+  set serialized (serialized) {
+    throw ImmutableError
+  }
 
-    stable.inplace(links, util.linkSort)
+  get size () {
+    return this.links.reduce((sum, l) => sum + l.size, this.serialized.length)
```
diasdavid: This is still getting calculated every single time size is requested, if a node has 100 links, this starts getting expensive.

src/index.js
diasdavid: sounds good

test/dag-node-test.js
```diff
@@ -29,10 +29,10 @@ module.exports = (repo) => {
         expect(Buffer.isBuffer(node.data)).to.be.true.mark()
         expect(node.size).to.be.above(0).mark()
 
-        util.serialize(node, (err, serialized) => {
+        DAGNode._serialize(node, (err, serialized) => {
```
diasdavid: the serialize and deserialize need to remain at the `util` package, it is the definition of the IPLD Format 

https://github.com/ipld/interface-ipld-format#ipld-format-utils

src/dag-node-factory.js
diasdavid: to ref: https://github.com/ipld/js-ipld-dag-pb/pull/7#issuecomment-260371998

> Note: It actualy might make sense to support also a promises API for the .create function since it is a top level consumed object (from ipfs.object.get), so that people don't have their promise chain broken, however, we are not dropping callbacks.

src/dag-node-factory.js
haadcode: I don't know, thought you left them there originally? :)

src/dag-node.js
haadcode: Pushed a change for this.

test/dag-node-test.js
haadcode: This doesn't make sense. Why would the serialization functionality be separate from all other functionality of a specific data structure (DAGNode)?

Also, not possible atm as deserialize uses .create() (two-way coupling between utils and dag-node-utils).

haadcode: Take that back, I understand what you mean with the comment. However, the util API doesn't make sense to me. Why would they need to be namespaced under utils instead of haing them on the top level? Ie.
```javascript
const DAGNode = require('dag-node')
const node = DAGNode.create(...)
DAGNode.serialize(node, ...)
DAGNode.deserialize(node, ...)
// etc.
```

haadcode: I can move the utils to one place, but my questions still stands: why do we need `utils` to be part of the API?

haadcode: Ignore all this, I'm not gonna bikeshed this now. I'll make some changes and push.

package.json
```diff
@@ -61,6 +61,7 @@
     "lodash": "^4.17.0",
     "ncp": "^2.0.0",
     "pre-commit": "^1.1.3",
-    "rimraf": "^2.5.4"
+    "rimraf": "^2.5.4",
+    "webpack": "2.1.0-beta.25"
```
dignifiedquire: not needed anymore

src/dag-link.js
```diff
@@ -2,28 +2,57 @@
 
 const mh = require('multihashes')
 
+const ImmutableError = new Error('Immutable property')
```
dignifiedquire: would be nicer to extend `Error` and use like this

```js
class ImmutableError extends Error {}

throw new ImmutableError('Size is immutable')
```

test/dag-node-test.js
diasdavid: @haadcode it does make sense because `util` and `resolver` is what an IPLD Resolver needs in order to process an IPLD Format, this is generic to any API resolver.

src/dag-node-utils.js
```diff
@@ -0,0 +1,210 @@
...
+function _cloneLinks (dagNode) {
+  return dagNode.links.length > 0 ? dagNode.links.slice() : []
+}
+
+exports.create = create
+exports.clone = clone
+exports.addLink = addLink
+exports.removeLink = removeLink
+exports.toDAGLink = toDAGLink
+exports.util = {
```
dignifiedquire: why are these namespaced and the others aren't?

package.json
diasdavid: same as https://github.com/libp2p/js-libp2p-floodsub/pull/9#discussion_r88196682

src/dag-node.js
dignifiedquire: If people want to shoot themselves in the foot they will find a way to do so ;)

src/dag-link.js
```diff
@@ -2,28 +2,57 @@
...
+  set size (size) {
+    throw ImmutableError
+  }
+
+  get multihash () {
+    return this._multihash
+  }
+
+  set multihash (multihash) {
+    throw ImmutableError
```
diasdavid: @dignifiedquire not sure if it applies to errors in the same way, but yeah, your suggestion above -- https://github.com/ipld/js-ipld-dag-pb/pull/7#discussion_r88208831 -- makes sense.

src/dag-node-utils.js
```diff
@@ -0,0 +1,210 @@
...
+  }
+  return data
+}
+
+function _cloneLinks (dagNode) {
+  return dagNode.links.length > 0 ? dagNode.links.slice() : []
+}
+
+exports.create = create
+exports.clone = clone
```
diasdavid: why did clone get separated from the class node? `const copy = node.clone()` was nice

src/dag-node-utils.js
```diff
@@ -0,0 +1,210 @@
...
+exports.addLink = addLink
```
diasdavid: still not convinced that these methods should not be at at the node level

src/dag-node-utils.js
haadcode: Same as for `node.addLink`. Let's either go with functional style, or OO style. Let's not mix the two.

src/dag-node-utils.js
haadcode: Of course they should. These are all functions that one can do to create and manipulate a DAGNode.

src/dag-node-utils.js
haadcode: See https://github.com/ipld/js-ipld-dag-pb/pull/7#pullrequestreview-8790101

README.md
```diff
@@ -50,9 +49,11 @@ dagPB.util
 Create a new DAGNode
 
 ```JavaScript
-var node = new dagPB.DAGNode([<data>, <[links]>])
+dagPB.DAGNode.create([<data>, <[links]>, hashAlgorithm, callback])
```
diasdavid: needs to be updated

README.md
```diff
@@ -50,9 +49,11 @@ dagPB.util
 Create a new DAGNode
 
 ```JavaScript
-var node = new dagPB.DAGNode([<data>, <[links]>])
+dagPB.DAGNode.create([<data>, <[links]>, hashAlgorithm, callback]) 
 ```
 
+#### TODO: update the functions docs from here on out
+
```
diasdavid: needs to be done

package.json
diasdavid: same as libp2p/js-libp2p-floodsub#9 (comment)

test/dag-node-test.js
haadcode: Why does IPLD need `util`? 

> this is generic to any API resolver.

What does that mean?

src/dag-link.js
```diff
@@ -5,25 +5,52 @@ const mh = require('multihashes')
...
+    return `DAGLink <${mhStr} - name: "${this.name}", size: ${this.size}>`
+  }
 
-    this.json = {
+  toJSON () {
+    return {
       name: this.name,
       size: this.size,
-      hash: this.multihash ? mh.toB58String(this.multihash) : undefined
+      hash: this.multihash ? mh.toB58String(this.multihash) : null
```
diasdavid: it is the same thing

src/dag-link.js
```diff
@@ -5,25 +5,52 @@ const mh = require('multihashes')
...
+  set size (size) {
+    throw new Error(`Can't set property: 'size' is immutable`)
+  }
+
+  get multihash () {
+    return this._multihash
+  }
+
+  set multihash (multihash) {
+    throw new Error(`Can't set property: 'multihash' is immutable`)
```
diasdavid: can we get that "ImmutableError" here too?

src/dag-node-utils.js
```diff
@@ -0,0 +1,210 @@
+'use strict'
```
diasdavid: if this is going to be the utils file, just keep the name of the file that was before

src/dag-node-utils.js
```diff
@@ -0,0 +1,210 @@
...
+  if ((nameOrLink.constructor &&
+       nameOrLink.constructor.name === 'DAGLink')) {
+    // It's a link
+    newLink = nameOrLink
+  } else if (typeof nameOrLink === 'string') {
+    // It's a name
+    if ((nodeOrMultihash.constructor &&
+       nodeOrMultihash.constructor.name === 'DAGNode')) {
+      // It's a node
+      // haadcode: not sure what ^ means, so the line below might not be correct
```
diasdavid: I had some `TODO` here for me with the steps I need to do in order to finish the function, why did you delete them?

diasdavid: See lines 67 to 97 https://github.com/ipld/js-ipld-dag-pb/pull/6/files#r88212711

src/dag-node-utils.js
```diff
@@ -0,0 +1,210 @@
...
+      newLink = new DAGLink(null, dagNode.size, nodeOrMultihash)
+    }
+  }
+
+  if (newLink) {
+    links.push(newLink)
+    sort(links, linkSort)
+  } else {
+    throw new Error('Link given as the argument is invalid')
+  }
```
diasdavid: This is incorrect, if the function has a callback, `error` should be pased on the callback.

src/dag-node-utils.js
```diff
@@ -0,0 +1,210 @@
...
+
+  create(data, links, callback)
+}
+
+function removeLink (dagNode, nameOrMultihash, callback) {
+  let data = _cloneData(dagNode)
```
diasdavid: can be a const

src/dag-node-utils.js
```diff
@@ -0,0 +1,210 @@
...
+  let links = _cloneLinks(dagNode)
```
diasdavid: can be const

src/dag-node-utils.js
```diff
@@ -0,0 +1,210 @@
...
+
+  if (typeof nameOrMultihash === 'string') {
+    links = dagNode.links.filter((link) => link.name !== nameOrMultihash)
+  } else if (Buffer.isBuffer(nameOrMultihash)) {
+    links = dagNode.links.filter((link) => !link.hash.equals(nameOrMultihash))
+  } else {
+    throw new Error('second arg needs to be a name or multihash')
```
diasdavid: if the function uses callbacks, errors should go in callbacks

package.json
haadcode: Installing with latest aegir still has the same problem as before: browser tests don't run. The error is
```
     Uncaught Error: Module '/Users/samuli/code/js-ipld-dag-pb/node_modules/buffer/index.js' is not a loader (must have normal or pitch function)
      at loadLoader (node_modules/loader-runner/lib/loadLoader.js:35:10)
      at iteratePitchingLoaders (node_modules/loader-runner/lib/LoaderRunner.js:164:2)
      at runLoaders (node_modules/loader-runner/lib/LoaderRunner.js:357:2)
      at NormalModule.doBuild (node_modules/webpack/lib/NormalModule.js:131:2)
      at NormalModule.build (node_modules/webpack/lib/NormalModule.js:179:15)
      at Compilation.buildModule (node_modules/webpack/lib/Compilation.js:127:9)
      at node_modules/webpack/lib/Compilation.js:306:10
      at node_modules/webpack/lib/NormalModuleFactory.js:74:13
      at NormalModuleFactory.applyPluginsAsyncWaterfall (node_modules/tapable/lib/Tapable.js:123:70)
      at onDoneResolving (node_modules/webpack/lib/NormalModuleFactory.js:49:11)
      at onDoneResolving (node_modules/webpack/lib/NormalModuleFactory.js:165:6)
      at node_modules/webpack/lib/NormalModuleFactory.js:161:6
      at node_modules/async/dist/async.js:3853:9
      at node_modules/async/dist/async.js:484:16
      at iteratorCallback (node_modules/async/dist/async.js:1084:13)
      at node_modules/async/dist/async.js:988:16
```

src/dag-node-utils.js
```diff
@@ -0,0 +1,210 @@
...
+  create(buf, links, callback)
+}
+
+function toProtoBuf (node) {
+  const pbn = {}
+
+  if (node.data && node.data.length > 0) {
+    pbn.Data = node.data
+  } else {
+    pbn.Data = null // new Buffer(0)
```
diasdavid: please don't change this, it needs to be a Buffer of size 0, we had a problem in the past because the serialized versions were different when data was empty (go vs js)

src/dag-node-utils.js
diasdavid: because of IPLD Format definition looking for the `util`

src/dag-node.js
diasdavid: fine

src/dag-node-utils.js
diasdavid: sounds good

src/dag-link.js
haadcode: Wrapping them to ImmutableError doesn't make sense anymore. What we would need to do is:
```javascript
class ImmutableError extends Error {
  constructor(error) {
    super(error)
  }
}
```

So let's just throw an Error.

src/dag-node-utils.js
haadcode: I didn't change this, it was like that in the original PR. Will change it to Buffer(0)

src/dag-link.js
haadcode: No, they're not. But let's not get into that now, will change it back to undefined.

src/dag-node-utils.js
haadcode: Actually, not gonna change this back. If I do, one of the tests starts failing:
```
  1) Node.js Tests DAGNode create an empty node:

      AssertionError: expected 2 to equal 0
      + expected - actual

      -2
      +0

      at DAGNode.create (test/dag-node-test.js:93:33)
```

Which way should it be? Change the code AND test, or leave it as it is?

src/utils.js
```diff
@@ -0,0 +1,206 @@
...
+    // It's a name
+    if ((nodeOrMultihash.constructor &&
+       nodeOrMultihash.constructor.name === 'DAGNode')) {
+      // It's a node
+      newLink = toDAGLink(nodeOrMultihash)
+    } else {
+      // It's a multihash
+      newLink = new DAGLink(null, dagNode.size, nodeOrMultihash)
+    }
+  }
```
diasdavid: This piece is incorrect, (also already gave reviews to this before). Again see line 67 to 97 https://github.com/ipld/js-ipld-dag-pb/pull/6#discussion_r88212711

Don't delete code + comments just because, information gets lost

Also, the addLink function, will need the `size` of link if just a name and multihash are passed.

src/utils.js
```diff
@@ -0,0 +1,206 @@
...
+
+  if (newLink) {
+    links.push(newLink)
+    sort(links, linkSort)
```
diasdavid: create does the sorting, we don't need to sort it twice

src/utils.js
```diff
@@ -0,0 +1,206 @@
...
+  } else {
+    return callback(new Error('Link given as the argument is invalid'), null)
+  }
+
+  create(data, links, callback)
+}
+
+function removeLink (dagNode, nameOrMultihash, callback) {
+  const data = _cloneData(dagNode)
+  let links = _cloneLinks(dagNode)
```
diasdavid: it can be const

src/utils.js
```diff
@@ -0,0 +1,206 @@
...
+}
+
+function deserialize (data, callback) {
+  const pbn = proto.PBNode.decode(data)
+
+  const links = pbn.Links.map((link) => {
+    return new DAGLink(link.Name, link.Tsize, link.Hash)
+  })
+
+  sort(links, linkSort)
```
diasdavid: create does sorting

src/utils.js
```diff
@@ -0,0 +1,206 @@
...
+  create(buf, links, callback)
+}
+
+function toProtoBuf (node) {
+  const pbn = {}
+
+  if (node.data && node.data.length > 0) {
+    pbn.Data = node.data
+  } else {
+    pbn.Data = null//new Buffer(0)
```
diasdavid: (also, already had reviewed) this is a bug that was already caught in the past, `null is not the same as new Buffer(0)` for protobuf serialization

diasdavid:
One more review. Thank you for enduring throughout this PR, it is really good to get you reviewing all of this as a consumer of all the API. Seems also that you are getting a speed up in IPLD from it too :)

Another comment: @haadcode I've reviewed at least 3 times the same 4 or 5 parts of this PR, some of which were bugs and still are there. It would save a lot of time if you make changes in smaller commits, avoid renaming files (util.js -> utils.js) and review every piece of feedback before asking for more, I feel a lot of information is getting lost and a lot of energy spent as well.

src/utils.js
haadcode: Yes and my question was: "Which way should it be?". Should I change the test so that it passed when this is changed to new Buffer(0)?

src/utils.js
haadcode: Please provide details as to *what* is incorrect when reviewing. Just saying this is incorrect is not helpful feedback. My original comment was "I don't understand what is meant here with your comments, please review if this is correct". Provide possible solutions, not just "this is not ok".

src/utils.js
haadcode: It can't, see the code couple of lines below.

src/utils.js
```diff
@@ -0,0 +1,206 @@
...
+
+const multihashing = require('multihashing-async')
+const CID = require('cids')
+const sort = require('stable').inplace
+const protobuf = require('protocol-buffers')
+const proto = protobuf(require('./dag.proto'))
+const DAGNode = require('./dag-node')
+const DAGLink = require('./dag-link')
+
+const hash = (type, data, cb) => multihashing(data, type, cb)
```
haadcode: This was changed in the code and accidentally left the line here.

haadcode:
> Another comment: @haadcode I've reviewed at least 3 times the same 4 or 5 parts of this PR, some of which were bugs and still are there. It would save a lot of time if you make changes in smaller commits, avoid renaming files (util.js -> utils.js) and review every piece of feedback before asking for more, I feel a lot of information is getting lost and a lot of energy spent as well.

Perhaps something went wrong in Github, but the "same 3 times" was with questions from me to you to clarify your comments. I have NOT committed anything without reviewing your feedback.

README.md
haadcode: What needs to be updated here? This is the function signature.

haadcode:
Ok, one more round. Addressed everything in the feedback expect promises. Everything else is "final" as far as I'm concerned.

haadcode:
After the CR and the required changes, I think the code is not as clean anymore as the original code in the PR was. 

Furthermore, the IPLD format api with the .resolver and .util (especially the util) doesn't make sense and should be reconsidered. Things are a) more coupled and b) don't provide clean namespacing/naming and structure and even makes the module internal implementation to adhere to those decisions.

However, for the sake of urgency, I'll leave this as is. There's been way too much bikeshedding regarding *code style* as oppose to functionality, correctness and API design and that takes away from motivation and from getting stuff done in timely manner. We need to stop using the argument "because it's like that elsewhere" as a catch-all when it clearly produces lower quality code and design.

Take what you want from this PR.

src/utils.js
diasdavid: It should be as it was, `new Buffer(0)`, this is not about our tests, it is about how protobuf serialization libraries work, we know for a fact that our 'empty objects' are different from go ones if we don't have a empty buffer when we go and serialize

src/utils.js
diasdavid: Doing push on array doesn't require it to be `let`

src/utils.js
diasdavid: You are correct, apologies for not being clear.

Right now, the thing that is missing from what I see is:
- [ ] We need to also pass the size of the link as an argument, because when we pass a name+multihash, we also need to pass the size of that link, otherwise the size of this node will be miscalculated
- [ ] This method can only be considered complete if the tests for adding and removing links are passing, right now they are commented out and with an incomplete migration, which is why I had this part in a TODO too. So, we need the tests to be migrated to this new API and making sure it is all correct

diasdavid:
> However, for the sake of urgency, I'll leave this as is. There's been way too much bikeshedding regarding code style as oppose to functionality

Sounds good 👍 



> Ok, one more round. Addressed everything in the feedback except promises. Everything else is "final" as far as I'm concerned.

This sound like functionality to me, what made you drop it? 

> https://github.com/ipld/js-ipld-dag-pb/pull/7#discussion_r88223470

This change will cause errors.

> https://github.com/ipld/js-ipld-dag-pb/pull/7#discussion_r88273246

❤️

Other minor things:
- [ ] https://github.com/ipld/js-ipld-dag-pb/pull/7#discussion_r88222980

diasdavid:
As discussed, I finished the last points of review in this PR, fixed tests, updated docs and also give a spin to a better addLink API

src/util.js
```diff
@@ -55,21 +25,42 @@ exports.serialize = (node, callback) => {
...
 
-exports.deserialize = (data, callback) => {
+function deserialize (data, callback) {
   const pbn = proto.PBNode.decode(data)
 
   const links = pbn.Links.map((link) => {
     return new DAGLink(link.Name, link.Tsize, link.Hash)
   })
 
-  stable.inplace(links, exports.linkSort)
```
diasdavid: *

PR #11, opened 2017-01-16
Converting string input data to Buffer.

JeffDownie:
Basically, I ran into this because I mis-read the README and started putting in string data into `DAGNode.create`! If you do this, the current behaviour allows a number of odd/confusing scenarios, and it's not clear that you've done something wrong.

Currently, if a string is put in as the data, it will be accepted, and will work when put or gotten from IPFS, but it will break if you try and clone the DAGNode due to the line `dagNode.data.copy(data)` (js strings don't have a copy method).

Passing in string data seemed like a reasonable thing to do, but I felt converting the input data string to a buffer would be better than allowing the internal DAGNode data to be a string - I felt having two internal data representations would perhaps cause other unknown problems, however, I am open to reverse this!

Also fixed a small typo in the README, because why not :)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,3 +35,6 @@ dist
 
 test/repo-just-for-test*
 /test/blocks
+
+# Vim .swp files
+**.swp
--- a/README.md
+++ b/README.md
@@ -71,7 +71,7 @@ const dagPB = require('ipld-dag-pb')
 dagPB.DAGNode.create  // create a DAGNode
 dagPB.DAGNode.clone   // clone a DAGNode
 dagPB.DAGNode.addLink // add a Link to a DAGNode, creating a new one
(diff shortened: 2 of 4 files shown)


JeffDownie:
I've added a check to ensure that data is a `Buffer` (after which argument is the callback has been worked out) - was that what you meant?

diasdavid:
@JeffDownie exactly :)

diasdavid:
added a notification for those following API changes here: https://github.com/ipfs/interface-ipfs-core/issues/55#issuecomment-272896106

Will give it a day before merging, if no one manifests anything against, of course :)

diasdavid:
Seems that no one raised any problem, going ahead and merging this. (sorry for having took so long!)

PR #25, opened 2017-02-26
bs58 is a normal dependecy

mitar:
It is used [here](https://github.com/ipld/js-ipld-dag-pb/blob/master/src/resolver.js#L4).


dignifiedquire:
Thank you :octocat:

PR #26, opened 2017-02-26
Be more robust when converting

mitar:


dignifiedquire:
Could you add a test that fails with the previous version, please?

mitar:
OK.

diasdavid:
Going to go ahead and merge this one. Agree with @dignifiedquire, it would be great to have a test, but since it is a one line change that won't provoke any new failures, I'll go ahead and unblock it.

mitar:
Oh, thanks.

PR #29, opened 2017-03-11
feat+fix: implement .tree and fix how the resolver was working

diasdavid:
In the process of implementing .tree, I've realized how we fell into the trap of 'transforming' the original data for convenience, for example, it was not possible to access the fields 'Name' or 'Tsize' of the protobuf, because we were handling the links as a direct jump to other objects.

This PR:

- [x] Fixes the resolver
- [x] Adds more tests
- [x] Implements tree
- [x] Implements isLink


diasdavid:
also make sure to check https://github.com/ipld/interface-ipld-format/pull/5

nicola:
sgtm

src/resolver.js
```diff
@@ -42,25 +39,36 @@ exports.resolve = (block, path, callback) => {
...
         const link = l.toJSON()
-        values[i] = link.multihash
+        values[i] = {
+          hash: link.multihash,
+          name: link.name,
+          size: link.size
+        }
+        // TODO by enabling something to resolve through link name, we are
+        // applying a transformation (a view) to the data, confirm if this
+        // is exactly what we want
```
nicola: is this a todo or a question? :)

src/resolver.js
```diff
@@ -42,25 +39,36 @@ exports.resolve = (block, path, callback) => {
       // for the resolver
       node.links.forEach((l, i) => {
         const link = l.toJSON()
-        values[i] = link.multihash
+        values[i] = {
+          hash: link.multihash,
+          name: link.name,
+          size: link.size
```
nicola: what is the size needed for?

src/resolver.js
```diff
@@ -1,10 +1,9 @@
 'use strict'
 
 const util = require('./util')
-const bs58 = require('bs58')
+// const bs58 = require('bs58')
```
dignifiedquire: please delete if you are not using it

src/resolver.js
```diff
@@ -42,25 +39,36 @@ exports.resolve = (block, path, callback) => {
...
+        value = { '/': value.size }
+      } else if (split[2] === 'Name') {
+        value = { '/': value.name }
       }
 
+      split.shift()
+      split.shift()
+      split.shift()
+
+      remainderPath = split.join('/')
```
dignifiedquire: these 4 lines could be simply

```js
remainderPath = split.slice(3).join('/')
```

src/resolver.js
```diff
@@ -78,26 +86,46 @@ exports.tree = (block, options, callback) => {
...
+    node.links.forEach((link, i) => {
+      paths.push(`Links/${i}/Name`)
+      paths.push(`Links/${i}/Tsize`)
+      paths.push(`Links/${i}/Hash`)
     })
 
-    if (node.data && node.data.length > 0) {
-      paths.push({ path: 'data', value: node.data })
-    }
+    paths.push('Data')
```
dignifiedquire: The above can be simplified to

```js
const links = node.links.map((l, i) => ([
  `Links/${i}/Name`,
  `Links/${i}/Tsize`,
  `Links/${i}/Hash`
])
links.push('Data')
const paths = Array.prototype.concat.apply(['Links'], links)
```

src/resolver.js
```diff
@@ -78,26 +86,46 @@ exports.tree = (block, options, callback) => {
...
 
     callback(null, paths)
   })
 }
+
+/*
+ * isLink: returns the Link if a given path in a block is a LInk, false otherwise
```
dignifiedquire: typo `LInk`

src/resolver.js
```diff
@@ -78,26 +86,46 @@ exports.tree = (block, options, callback) => {
...
+    if (err) {
+      return callback(err)
+    }
+
+    if (result.remainderPath.length > 0) {
+      return callback(new Error('path out of scope'))
+    }
+
+    if (typeof result.value === 'object' && result.value['/']) {
+      // result.value['/'] = bs58.encode(result.value['/'])
```
dignifiedquire: please remove if not used

src/resolver.js
diasdavid: `TODO: Ask the question:....` :D

src/resolver.js
diasdavid: It is the size of this branch of the graph so far. It is needed to avoid traversing graphs over and over. go-ipfs has it since forever.

src/resolver.js
diasdavid: 👍

PR #32, opened 2017-03-17
feat: upgrade to new ipfs-block and blockservice

dignifiedquire:
Ref https://github.com/ipfs/js-ipfs/issues/787


test/dag-node-test.js
```diff
@@ -406,51 +411,29 @@ module.exports = (repo) => {
...
           }
         ], done)
       })
     })
 
     it('deserialize go-ipfs block from ipldResolver', (done) => {
-      const bs = new BlockService(repo)
+      if (!isNode) {
+        return done()
+      }
```
diasdavid: Why remove this test from the browser test suit?

test/browser.js
```diff
@@ -1,58 +1,42 @@
...
-      )
-    }), done)
+    ], done)
   })
 
-  const repo = new IPFSRepo(path, {stores: Store})
-
   require('./dag-node-test')(repo)
   require('./dag-link-test')(repo)
 })
```
diasdavid: This browser.js was also setting up the repo with some blocks for tests.

dignifiedquire: Remember when I said we turn sharding off in the browser? That's the reason that setup doesn't work anymore.

test/dag-node-test.js
dignifiedquire: See below

test/browser.js
diasdavid: roger :)

PR #51, opened 2018-01-03
Fix false positives in isLink and update test

AdamStone:
The isLink method (formerly isCID) appears [intended](https://github.com/ipld/js-ipld-dag-pb/blob/master/test/resolver.spec.js#L222) to identify paths specifically to link hashes (as opposed to paths to link name or size), but currently falls into the `true` case for paths to any truthy values. The tests pass because the non-valid CID case happens to check a path to an empty value.

Discussion [here](https://github.com/ipfs/js-ipfs/pull/1045#discussion_r159034560).

### DIFF
--- a/test/resolver.spec.js
+++ b/test/resolver.spec.js
@@ -228,10 +228,16 @@ describe('IPLD Format resolver (local)', () => {
   })
 
   it('resolver.isLink for non valid CID', (done) => {
-    resolver.isLink(dataLinksNodeBlock, 'Links/0/Tsize', (err, link) => {
+    // blank value case
+    resolver.isLink(dataLinksNodeBlock, 'Links/0/Name', (err, link) => {
       expect(err).to.not.exist()
       expect(link).to.equal(false)
-      done()
+      // non-blank value case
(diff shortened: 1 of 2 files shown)


test/resolver.spec.js
```diff
@@ -228,7 +228,7 @@ describe('IPLD Format resolver (local)', () => {
   })
 
   it('resolver.isLink for non valid CID', (done) => {
-    resolver.isLink(dataLinksNodeBlock, 'Links/0/Name', (err, link) => {
+    resolver.isLink(dataLinksNodeBlock, 'Links/0/Tsize', (err, link) => {
```
diasdavid: Why the change in the test? It should also be false for /Name.

AdamStone: True, but maybe it should check both then. Checking only name doesn't cover the general (non-blank-value) case.

AdamStone:
@diasdavid I updated the test to check both Name and Size. I linked it and ran the ipld-resolver tests as well, and they also pass, but it seems like they don't actually hit this function.

PR #58, opened 2018-03-08
replace constructor.name with instanceof so uglify can mangle and compress

ya7ya:
this is part of ongoing effort to properly fix https://github.com/ipfs/js-ipfs/issues/1131

it replaces `link.constructor.name === 'DAGLink'` with a friendlier `link instanceof DAGLink`


dignifiedquire:
please revert this, there is a reason we don‘t use instanceof anywhere. It breaks for consumers using the library if they bundle a different set of code or the class.

PR #68, opened 2018-04-25
fix: Initialise the DAGLink name to empty string if a falsey value is passed

achingbrain:
Changes the behaviour of `new DAGLink(...)` to default the link name to an empty string, otherwise you can get odd behaviour as detailed in ipld/js-ipld#125 where you fetch a `DAGNode` using a `CID` and it is returned with a different `CID` to the one you used to fetch it..

### DIFF
--- a/src/dag-link/index.js
+++ b/src/dag-link/index.js
@@ -12,7 +12,7 @@ class DAGLink {
     //  note - links should include size, but this assert is disabled
     //  for now to maintain consistency with go-ipfs pinset
 
-    this._name = name
+    this._name = name || ''
     this._size = size
 
     if (typeof multihash === 'string') {
--- a/test/dag-node-test.js
+++ b/test/dag-node-test.js
@@ -134,6 +134,29 @@ module.exports = (repo) => {
       })
     })
 
+    it('create with undefined link name', (done) => {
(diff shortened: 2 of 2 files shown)


vmx:
Have you checked what the Go implementation is doing when Go's notion of empty/undefined is passed in? I'm just asking as I'd be interested and if you haven't find it out myself :)

achingbrain:
On the CLI at least it looks like falsey `Name`s get changed to empty strings:

```
$ echo '{"Data":"Hello world","Links":[{"Hash":"QmUxD5gZfKzm8UN4WaguAMAZjw2TzZ2ZUmcqm2qXPtais7","Name":null,"Size":819}]}' | ipfs object put
added Qma1CbV7jPXncZjNC5jgbHs69ggbrfTvT5V8XhyiM4zvVp
$ ipfs dag get Qma1CbV7jPXncZjNC5jgbHs69ggbrfTvT5V8XhyiM4zvVp
{"data":"SGVsbG8gd29ybGQ=","links":[{"Name":"","Size":819,"Cid":{"/":"QmUxD5gZfKzm8UN4WaguAMAZjw2TzZ2ZUmcqm2qXPtais7"}}]}
```

PR #66, opened 2018-04-16
fix: lock stable dep

fsdiogo:
This PR locks the `stable` dependecy to a working version, as the latest one breaks all the browser tests.

Fixes https://github.com/ipld/js-ipld-dag-pb/issues/65.

PR #69, opened 2018-04-26
fix: unlock and update stable dep

fsdiogo:
A new version of stable has been released that fixes an error that was breaking the browser tests.
So we can now unlock the stable dependency and update its version.


fsdiogo:
For cross reference, check https://github.com/ipld/js-ipld-dag-pb/pull/66.

PR #74, opened 2018-06-23
feat: add util.cid options

richardschneider:
See https://github.com/ipld/interface-ipld-format/issues/40

### DIFF
--- a/src/util.js
+++ b/src/util.js
@@ -3,16 +3,42 @@
 const CID = require('cids')
 const protons = require('protons')
 const proto = protons(require('./dag.proto.js'))
+const resolver = require('./resolver')
 const DAGLink = require('./dag-link')
 const DAGNode = require('./dag-node')
+const multihashing = require('multihashing-async')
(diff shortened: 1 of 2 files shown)


src/util.js
```diff
@@ -3,16 +3,42 @@
...
+ *
+ * @param {Object} dagNode - Internal representation
+ * @param {Object} [options] - Options to create the CID
+ * @param {number} [options.version] - CID version number. Defaults to zero if hashAlg == 'sha2-256'; otherwise, 1.
+ * @param {string} [options.hashAlg] - Defaults to hashAlg for the resolver
+ * @param {CidCallback} callback - Callback that handles the return value
+ * @returns {void}
+ */
+function cid (dagNode, options, callback) {
+  if (options instanceof Function) {
```
vmx: Searching through the existing IPFS/IPLD code base it seems that the preferred way for checking if something is a function is: ` if (typeof options !== 'function') {`. My guess is that it is related to code minifiers. Please change it.

PR #75, opened 2018-06-26
fix: pass serialized blob to util.cid

richardschneider:
BREAKING CHANGE: the first argument is now the serialized output NOT the dag node.
See https://github.com/ipld/interface-ipld-format/issues/32

### DIFF
--- a/src/util.js
+++ b/src/util.js
@@ -18,9 +18,9 @@ exports = module.exports
  * @param {?CID} cid - CID if call was successful
  */
 /**
- * Get the CID of the DAG-Node.
+ * Get the CID of the serialized ProtoBuf node.
  *
- * @param {Buffer} blob - Serialized binary data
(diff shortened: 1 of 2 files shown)


src/util.js
```diff
@@ -19,14 +20,15 @@ exports = module.exports
 /**
  * Get the CID of the DAG-Node.
  *
- * @param {Object} dagNode - Internal representation
+ * @param {Buffer} blob - Serialized binary data
```
vmx: I would change it to `Encoded Protocol Buffers blob` or `Serialized as Protocol Buffers`.

src/util.js
```diff
@@ -19,14 +20,15 @@ exports = module.exports
 /**
  * Get the CID of the DAG-Node.
```
vmx: Perhaps it should say `Get CID of Protocol Buffers encoded node`?

PR #78, opened 2018-07-19
fix: add support for resolving links by name

diasdavid:
This is just a stab at fixing the _resolve links by name_ on dag-pb nodes. This PR is not finished (at all) but it shows what is missing.

Short story: go-ipfs and js-ipfs historically have supported the ability to do `jsipfs dag get CID/a` where a is a named link in a dag-pb.

So a dag-pb node like this.

```
{
  Data: <some data>
  Links: [
    {
       Name: a
       Hash: b58...
       Size: 1000
    }
  ]
}
```

Should support resolving the link through name so that both `jsipfs dag get CID/a` and `jsipfs dag get CID/Links/0/Hash` are equivalent.

### DIFF
--- a/src/resolver.js
+++ b/src/resolver.js
@@ -57,6 +57,8 @@ exports.resolve = (binaryBlob, path, callback) => {
           values[link.name] = link.multihash
         })
 
+        console.log(values)
+
         let value = values[split[1]]
 
         // if remainderPath exists, value needs to be CID
@@ -74,7 +76,33 @@ exports.resolve = (binaryBlob, path, callback) => {
       } else if (split[0] === 'Data') {
         cb(null, { value: node.data, remainderPath: '' })
       } else {
-        cb(new Error('path not available'))
+        const values = {}
+
+        // populate both index number and name to enable both cases
+        // for the resolver
+        node.links.forEach((l, i) => {
+          const link = l.toJSON()
+          values[i] = {
+            hash: link.multihash,
+            name: link.name,
+            size: link.size
+          }
+          // TODO by enabling something to resolve through link name, we are
+          // applying a transformation (a view) to the data, confirm if this
+          // is exactly what we want
+          values[link.name] = link.multihash
+        })
+
+        const val = values[split[0]]
+
+        if (val) {
+          return cb(null, {
+            value: { '/': val.hash },
+            remainderPath: split.slice(3).join('/')
+          })
+        } else {
+          cb(new Error('path not available'))
+        }
       }
     }
   ], callback)
--- a/src/resolver.js
+++ b/src/resolver.js
@@ -46,9 +46,6 @@ exports.resolve = (binaryBlob, path, callback) => {
         // for the resolver
         node.links.forEach((l, i) => {
           const link = l.toJSON()
-          // TODO by enabling something to resolve through link name, we are
-          // applying a transformation (a view) to the data, confirm if this
-          // is exactly what we want
           values[i] = values[link.name] = {
             hash: link.multihash,
             name: link.name,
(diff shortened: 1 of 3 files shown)


diasdavid:
@vmx can you check on this with @olizilla and @alanshaw. They are looking for this feature in for the WebUI.

@VictorBjelkholm can we make sure to disable codecov comments?

alanshaw:
I hope you don't mind, I made some fixes and added a test - should be good to go now.

src/resolver.js
```diff
@@ -46,15 +46,14 @@ exports.resolve = (binaryBlob, path, callback) => {
         // for the resolver
         node.links.forEach((l, i) => {
           const link = l.toJSON()
-          values[i] = {
+          // TODO by enabling something to resolve through link name, we are
+          // applying a transformation (a view) to the data, confirm if this
+          // is exactly what we want
+          values[i] = values[link.name] = {
```
vmx: I don't think `values[link.name]` is needed here. The case is handled below.

Please also remove the "TODO".

src/resolver.js
```diff
@@ -74,6 +73,31 @@ exports.resolve = (binaryBlob, path, callback) => {
       } else if (split[0] === 'Data') {
         cb(null, { value: node.data, remainderPath: '' })
       } else {
+        const values = {}
```
vmx: Can you please add a comment saying that this is the case of a named link? It took me a while to figure that out.

src/resolver.js
```diff
@@ -74,6 +73,31 @@ exports.resolve = (binaryBlob, path, callback) => {
...
+
+        // populate both index number and name to enable both cases
+        // for the resolver
+        node.links.forEach((l, i) => {
+          const link = l.toJSON()
+          // TODO by enabling something to resolve through link name, we are
+          // applying a transformation (a view) to the data, confirm if this
+          // is exactly what we want
+          values[i] = values[link.name] = {
```
vmx: The assignment to `values[i]` is not needed here and I find it rather confusing.

Please remote the comments/change them to make the saying the right thing.

test/resolver.spec.js
```diff
@@ -146,6 +146,15 @@ describe('IPLD Format resolver (local)', () => {
...
 
+      it('links by name', (done) => {
+        resolver.resolve(linksNodeBlob, 'named link', (err, result) => {
+          expect(err).to.not.exist()
+          expect(result.value['/']).to.eql(links[1].multihash)
+          expect(result.remainderPath).to.eql('')
+          done()
+        })
+      })
+
```
vmx: Could you please add tests for:
 - named link and a remainder path
 - error case for non-existent named link

src/resolver.js
alanshaw: This was already in the code base although I don't think it was working correctly:
https://github.com/ipld/js-ipld-dag-pb/blob/fda23c444b5437fe15c2875dfe547ad1e63a2c27/src/resolver.js#L54-L57

It allows you to reference links by name via Links e.g. `/ipfs/QmHash/Links/myNamedLink`

vmx: Got it. So this is kind of the IPFS vs. IPLD like traversal.

alanshaw: Yeah, I don't know if it's _really_ necessary to be able to `/ipfs/QmHash/Links/myNamedLink` **as well as** `/ipfs/QmHash/myNamedLink` but the former was already there and we're adding the latter.

vmx: `/ipfs/QmHash/Links/myNamedLink` is what I call IPLD-like traversal. It will then be use if you use the `/ipld` prefix.

alanshaw:
Is good now @vmx? macOS tests are npm failure not test failure.

vmx:
@alanshaw yes it's good to be merged.

PR #81, opened 2018-08-02
perf: make this._json calculation lazy

achingbrain:
I've been doing some profiling of js-ipfs and where there are large amounts of DAG operations being performed (importing all of npm to ipfs, for example) we spend about 1.8-2.5% of our time in the creation of this._json objects, even though they are not always used, so this PR makes them be
created when first requested.

It also passes the `this._json` object through `Object.freeze` because otherwise we expose a mutable object to the world - it seems odd that we've taken pains to ensure that the properties of a DAGNode/DAGLink are immutable, then expose objects whose properties are not immutable.

Finally it makes more use of the `cids` module to parse CIDs from multihashes as the `cids` module respects CIDv1 versions/codecs if present whereas the `multihashes` module does not and is only really suitable for use with CIDv0.

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,7 +61,6 @@
     "cids": "~0.5.3",
     "class-is": "^1.1.0",
     "is-ipfs": "~0.4.2",
-    "multihashes": "~0.4.13",
     "multihashing-async": "~0.5.1",
     "protons": "^1.0.1",
     "pull-stream": "^3.6.8",
@@ -78,6 +77,7 @@
     "ipfs-block-service": "~0.14.0",
     "ipfs-repo": "~0.22.1",
     "lodash": "^4.17.10",
+    "multihashes": "~0.4.13",
     "ncp": "^2.0.0",
     "rimraf": "^2.6.2"
   }
--- a/src/dag-link/index.js
+++ b/src/dag-link/index.js
@@ -14,27 +14,23 @@ class DAGLink {
 
     this._name = name || ''
     this._size = size
-
-    if (typeof multihash === 'string') {
-      const cid = new CID(multihash)
-      this._multihash = cid.buffer
-    } else if (Buffer.isBuffer(multihash)) {
-      this._multihash = multihash
-    }
+    this._cid = new CID(multihash)
   }
 
   toString () {
-    const cid = new CID(this.multihash)
-    return `DAGLink <${cid.toBaseEncodedString()} - name: "${this.name}", size: ${this.size}>`
+    return `DAGLink <${this._cid.toBaseEncodedString()} - name: "${this.name}", size: ${this.size}>`
   }
 
   toJSON () {
-    const cid = new CID(this.multihash)
-    return {
-      name: this.name,
-      size: this.size,
-      multihash: cid.toBaseEncodedString()
+    if (!this._json) {
+      this._json = Object.freeze({
+        name: this.name,
+        size: this.size,
+        multihash: this._cid.toBaseEncodedString()
+      })
     }
+
+    return this._json
   }
 
   get name () {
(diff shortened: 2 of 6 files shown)


package.json
```diff
@@ -78,6 +77,7 @@
     "ipfs-block-service": "~0.14.0",
     "ipfs-repo": "~0.22.1",
     "lodash": "^4.17.10",
+    "multihashes": "~0.4.13",
```
olizilla: I think `multihashes` is no needed after this very good change.

olizilla:
@achingbrain @vmx @diasdavid as per https://github.com/ipld/js-ipld-dag-pb/pull/80#issue-205188915 are there any objections to providing a `cid` getter on `DagLink` and `DagNode`? 

Ideally we'd remove the `mutlihash` property and release as a major version, having both props and a deprecation warning seems more mess than helpful, though I could be wrong.

I'm already normalising all IPLD nodes to have a `cid` property in IPLD explorer https://github.com/ipfs-shipyard/ipld-explorer/blob/14fee7a65773b32c78c82ff58ce99c70c9d28a81/src/lib/dag.js#L27-L51

package.json
achingbrain: It's [still being used in a test](https://github.com/ipld/js-ipld-dag-pb/blob/master/test/dag-node-test.js#L432) which is why it's moved from `dependencies` to `devDependencies`.

olizilla:
I had imagined it'd return the buffer to be a more minimal change. In the common case of cidv0 it'd mean the usage site just has to rename the property.

I do have to upgrade the buffers to CID objects all over the place right now, so it'd help to use CID objects over buffers in the object representations, but we would have to review all the IPLD node types. 

I think we should do the smaller change now... renaming multihash to cid but have it still return a Buffer. Then as a more awesome endeavour, we should figure out if all links in all nodes should be represented as a CID object rather than a Buffer.

achingbrain:
I think it’d be better to just have .cid return the object as, like you say, people convert them to objects anyway, particularly if you then want to work with the ipld APIs at all, which you probably want to if you are messing with CIDs in the first place.

I’d leave .multihash returning a Buffer in place for backwards compatibility and perhaps consider getting it to print a warning and removing it at some point in future.

Otherwise we make one breaking change to remove .multihash and add .cid, then potentially another to go from Buffer to Object.  We can swerve all that disruption by just going for the useful API straight off.

What do you think?

olizilla:
Compelling argument @achingbrain. I don't have a feel for the direction the IPLD interfaces are supposed to be moving in, I had assumed that using Buffers all over the place was deliberate and I just handn't found the repo that explained it yet, but I do agree it doesn't make for an intuitive API for people want to traverse IPLD nodes.

Vaguely related to this is the discussion around what is the standard pattern for the IPLD transform that we apply when representing entities from other systems as objects https://github.com/ipld/js-ipld-ethereum/issues/25

vmx:
@achingbrain @olizilla It's funny that the CID object vs. Buffer comes up as @mikeal and I were discussing the same thing during the DWeb Summit. Combining that discussion with yours above I'd say the outcome would be that we have a new `cid` field which contains the CID object.

Please everyone mentioned in this comment give a thumb up if you agree, or post a comment if you don't. It'll then be a separate change, independent of this one.

achingbrain:
This is the sort of thing I am seeing while importing npm to IPFS without this PR:

image

The other really long flat bar is the garbage collector.

Flame graph from [appmetrics-dash](https://www.npmjs.com/package/appmetrics-dash).

achingbrain:
Can it be merged & released?

vmx:
@achingbrain Sorry this somehow slipped my attention. I've also done a release.

PR #83, opened 2018-08-13
fix: allow mutating returned .toJSON value

achingbrain:
Relaxes the constraints from ipld/js-ipld-dag-pb#81 a little as it's caused quite a lot of tests to break.

We still have an immutable copy in the object but any objects returned will be mutable.


vmx:
Commit linting only breaks due to no disk space, that should block the merge.

PR #89, opened 2018-09-26
chore: remove CI files

vmx:
The only CI we are using is Jenkins, remove files from all other services.


vmx:
CI passed, it just didn't update the results as it was a "rebuild" and not a "retry": https://ci.ipfs.team/blue/organizations/jenkins/IPLD%2Fjs-ipld-dag-pb/detail/PR-89/2/pipeline

PR #92, opened 2018-10-09
feat: expose cid property on DAGLinks and DAGNodes

achingbrain:
Implements the change proposed in https://github.com/ipld/js-ipld-dag-pb/pull/81#issuecomment-410681495


achingbrain:
Any chance of a release with this in?

PR #98, opened 2018-10-30
perf: fixes #97 by not sorting DAGNode links when they are already sorted

### LINKED ISSUE #97: Skip sorting links when deserializing?
At the moment when deserializing a protobuf we [pass](https://github.com/ipld/js-ipld-dag-pb/blob/master/src/util.js#L75) the links into `DAGNode.create` which in turn [passes](https://github.com/ipld/js-ipld-dag-pb/blob/master/src/dag-node/create.js#L39) them to [linkSort](https://github.com/ipld/js-ipld-dag-pb/blob/master/src/dag-node/util.js#L24-L29).

When you have DAGNodes with lots of links being queried repeatedly this can take up a non-trivial amount of time.

I think we should be able to skip the sorting.  My reasoning is:

1. We're deserializing the DAGNode from the buffer that corresponds to it's CID, so the links should have been sorted already
2. If the links were not sorted, serializing the DAGNode with sorted links would result in one with a different CID as the serialized bytes would be in a different order

Any thoughts @ipld/javascript-team ?

achingbrain:
See #97 for context.  Essentially when we are deserializing a DAGNode from a buffer, the links should already be sorted so just new up a DAGNode object and return it.  Though we still have to calculate what the CID was that *may* have been used to load this DAGNode.  See ipld/js-ipld#173 for further discussion on that point.

### DIFF
--- a/src/util.js
+++ b/src/util.js
@@ -72,7 +72,25 @@ function deserialize (data, callback) {
 
   const buf = pbn.Data == null ? Buffer.alloc(0) : Buffer.from(pbn.Data)
 
-  DAGNode.create(buf, links, callback)
+  // Work out the CID that was used to load this node
+  // Will be inaccurate if the hash-alg was not the default sha2-256
+  // Should be able to go away when https://github.com/ipld/js-ipld/issues/173 is resolved
+  serialize({
+    data: buf,
+    links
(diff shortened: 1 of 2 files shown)


achingbrain:
@vmx have updated the commit message

PR #99, opened 2018-10-31
BREAKING CHANGE: Remove .cid, .multihash and .serialized properties

achingbrain:
Follows on from https://github.com/ipld/js-ipld/issues/173#issuecomment-434408680

These properties are removed from the DAGNode class.

* `.multihash` is removed because they aren't multihashes any more
* `.cid` is removed to bring dag-pb in line with other ipld types
* `.serialized` is removed because storing data buffers and the serialized form uses too much memory - we can use the utils.serialize method to create the serialized form when we need it, which in this module is just during the tests

`.multihash` has also changed to `.cid` in the output of `DAGLink.toJSON` and `DAGNode.toJSON` because since CIDv1 they are not just multihashes any more; the multihash is contained within the CID.

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,7 +178,7 @@ Creates a new DAGNode instance with the union of node.links plus the new link.
 {
   name: '<some string>', // optional
   size: <size in bytes>,
-  cid: <cid> // can be a String CID or CID buffer
+  cid: <cid> // can be a String CID, CID buffer or CID object
 }
 ```
 
@@ -186,7 +186,7 @@ Creates a new DAGNode instance with the union of node.links plus the new link.
 #### rmLink(node, nameOrCid, callback)
 
 - `node` - type: DAGNode
-- `nameOrCid` - type: String or CID buffer
+- `nameOrCid` - type: String, CID object or CID buffer
 - `callback` - type: function with signature `function (err, node) {}`
 
 Removes a link from the node by name. Returns a *new* instance of DAGNode without modifying the old one.
@@ -242,7 +242,7 @@ const DAGLink = dagPB.DAGLink
 DAGLink.create(
   'link-to-file',  // name of the link (can be empty)
   10,              // size in bytes
-  'QmSomeHash...', // can be cid buffer or string
+  'QmSomeHash...', // can be CID object, CID buffer or string
   (err, link) => {
     if (err) {
       throw err
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@
     - [Add and remove a Link](#add-and-remove-a-link)
 - [API](#api)
   - [DAGNode functions](#dagnode-functions)
-    - [DAGNode.create(data, links, hashAlg, callback)](#dagnodecreatedata-links-hashalg-callback)
+    - [DAGNode.create(data, links, callback)](#dagnodecreatedata-links-callback)
     - [addLink(node, link, callback)](#addlinknode-link-callback)
     - [rmLink(node, nameOrCid, callback)](#rmlinknode-nameorcid-callback)
     - [clone(node, callback)](#clonenode-callback)
@@ -134,11 +134,10 @@ const dagPB = require('ipld-dag-pb')
 const DAGNode = dagPB.DAGNode
 ```
 
-#### DAGNode.create(data, links, hashAlg, callback)
+#### DAGNode.create(data, links, callback)
 
 - `data` - type: Buffer
 - `links`- type: Array of DAGLink instances or Array of DAGLink instances in its json format (link.toJSON)
-- `hashAlg` - type: String
 - `callback` - type: function with signature `function (err, node) {}`
 
 Create a DAGNode.
@@ -181,7 +180,6 @@ Creates a new DAGNode instance with the union of node.links plus the new link.
   cid: <cid> // can be a String CID, CID buffer or CID object
 }
 ```
-
 
 #### rmLink(node, nameOrCid, callback)
 
--- a/README.md
+++ b/README.md
@@ -28,22 +28,20 @@
   - [DAGNode functions](#dagnode-functions)
     - [DAGNode.create(data, links, hashAlg, callback)](#dagnodecreatedata-links-hashalg-callback)
     - [addLink(node, link, callback)](#addlinknode-link-callback)
-    - [rmLink(node, nameOrMultihash, callback)](#rmlinknode-nameormultihash-callback)
+    - [rmLink(node, nameOrCid, callback)](#rmlinknode-nameorcid-callback)
     - [clone(node, callback)](#clonenode-callback)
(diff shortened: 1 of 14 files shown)


vmx:
First quick comment (as gatekeeper of good/correct commit messages). This won't pass commit lint. According to our style it would be:

```
feat: remove .cid, .multihash and .serialized properties

Follows on from https://github.com/ipld/js-ipld/issues/173#issuecomment-434408680

BREAKING CHANGE:
These properties are removed from the DAGNode class.

* `.multihash` is removed because they aren't multihashes any more
* `.cid` is removed to bring dag-pb in line with other ipld types
* `.serialized` is removed because storing data buffers and the
  serialized form uses too much memory - we can use the utils.serialize
  method to create the serialized form when we need it, which in this
  module is just during the tests

`.multihash` has also changed to `.cid` in the output of
`DAGLink.toJSON` and `DAGNode.toJSON` because since CIDv1 they are
not just multihashes any more; the multihash is contained within
the CID.
```

achingbrain:
It would be nice if we could discover these things before CI gets involved.  I'd added commit linting to aegir, but you can only lint commits once they've been committed (I think), so it had to be part of a pre-push hook, which we removed from all of our repos a little while back. 🤷‍♂️

vmx:
This repo (and all other IPLD repos) still have a pre-push hook :)

achingbrain:
Neat, well you can add [`aegir lint-commits`](https://github.com/ipfs/aegir/blob/master/cmds/lint-commits.js) to them all like how [this used to be](https://github.com/ipfs/js-ipfs-mfs/commit/5322d73af927245fe0554b43ed3eb963ae05377c#diff-b9cfc7f2cdf78a7f4b91a753d10865a2)..

achingbrain:
Cool, well you can add [`aegir lint-commits`](https://github.com/ipfs/aegir/commit/0597098988be025a0fd3b26bafa173b4326652d4) to them and never be surprised by CI ever again.  Well, not for commit messages anyway.

achingbrain:
We are way off topic.

README.md
```diff
@@ -180,21 +178,21 @@ Creates a new DAGNode instance with the union of node.links plus the new link.
...
 }
 ```
 
 
-#### rmLink(node, nameOrMultihash, callback)
+#### rmLink(node, nameOrCid, callback)
 
 - `node` - type: DAGNode
-- `nameOrMultihash` - type: String or multihash buffer
+- `nameOrCid` - type: String or CID buffer
```
vmx: It can also be a CID object instance.

README.md
```diff
@@ -242,13 +236,13 @@ const dagPB = require('ipld-dag-pb')
...
 
-#### DAGLink.create(name, size, multihash, callback)
+#### DAGLink.create(name, size, cid, callback)
 
 ```JavaScript
 DAGLink.create(
   'link-to-file',  // name of the link (can be empty)
   10,              // size in bytes
-  'QmSomeHash...', // can be multihash buffer or string
+  'QmSomeHash...', // can be cid buffer or string
```
vmx: Or a CID object instance.

achingbrain:
@vmx yes, please wait and do other releases in the meantime - if for nothing else other than I'm getting PRs ready for unixfs and mfs that will fix the breakage this will cause.

achingbrain:
How about removing the `size` property too?  It's the size of the serialized buffer that a `DAGNode` would be turned into by [`util.serialize`](https://github.com/ipld/interface-ipld-format#utilserializedagnode-callback) and doesn't seem to be the sort of thing that a data object would know about.  It also means we don't have to calculate the serialized size on creation so we'd save a bunch of CPU cycles there too.

achingbrain:
This does make calculating the return values from `ipfs.files.add` difficult as it contains the byte size of the `DAGNode` tree that stores a given file (e.g. including protobuf wrappers, serialized DAGLinks, etc).

Is that information actually useful to anyone though?  I remember my surprise when the size returned from `ipfs.files.add` was not the size of the file I'd passed in.

achingbrain:
I'm going to back out the removal of `.size` for the time being until we've got some numbers on it being a bottleneck.

achingbrain:
I've opened ipfs/interface-ipfs-core#387 to talk about the `size` property returned from `ipfs.files.add` a bit more.

diasdavid:
@achingbrain ack on https://github.com/ipfs/interface-ipfs-core/issues/387. Do you plan to block this PR on that issue? I recommend that we make that change as separate.

achingbrain:
@diasdavid - I agree, that issue should not block this PR, we can make the change later once we've got consensus on whether to make the API change or not.

diasdavid:
IIUC, this is a big perf improve. Tagging it as P1, can we get this done asap?

achingbrain:
This one should be good to merge & release - I think all the PRs for everything else are in place, they'll just need the GH URLs removing from the various `package.json` changes and deps updating as we go along.

test/dag-node-test.js
```diff
@@ -282,14 +272,11 @@ module.exports = (repo) => {
             cb()
           })
         },
-        (cb) => {
-          const link = toDAGLink(node2).toJSON()
-
+        (cb) => toDAGLink(node2, cb),
```
vmx: The `toJSON()` is missing. The test is about testing `addLink()` *by object*.

src/dag-node/create.js
```diff
@@ -1,57 +1,43 @@
...
+const {
+  serialize
+} = require('../util.js')
 const dagNodeUtil = require('./util.js')
 const linkSort = dagNodeUtil.linkSort
 const DAGNode = require('./index.js')
 const DAGLink = require('../dag-link')
 
-function create (data, dagLinks, hashAlg, callback) {
+function create (data, links, callback) {
```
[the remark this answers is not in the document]
vmx: TBH, I don't quite follow as even a ipld.dag.put would need to forward it to dag-pb at some point. But it sounds like you have it under control.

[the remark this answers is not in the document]
vmx: I wait, I got it. You pass in the already encoded data.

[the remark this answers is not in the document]
achingbrain: It does - `ipld.dag.put` passes it to [`util.cid`](https://github.com/ipld/js-ipld-dag-pb/blob/master/src/util.js#L29-L45).

test/dag-node-test.js
```diff
@@ -282,14 +272,11 @@ module.exports = (repo) => {
             cb()
           })
         },
-        (cb) => {
-          const link = toDAGLink(node2).toJSON()
-
+        (cb) => toDAGLink(node2.toJSON(), cb),
```
vmx: This should be 

    (cb) => toDAGLink(node2, cb).toJSON()

PR #102, opened 2018-11-12
fix: reinstate DAGLink.multihash property with deprecation notice

achingbrain:
Just having a slight wobble about removing the `.multihash` property from the DAGLink class as part of #99.  It's the right thing to do but it will cause a lot of disruption.


achingbrain:
Feel free to call me a coward and close this without merging.

PR #105, opened 2018-12-06
WIP: new ipld format api

vmx:
This is an implementation of the [new IPLD Format API](https://github.com/ipld/interface-ipld-format/pull/50). It hopefully shows that although the public API changes a lot, the changes make sense and it won't be too hard to upgrade.

The resolver is completely gone, all enumerable properties are the public API to resolve through now. See the `resolver.spec.js` test for how things work now.


achingbrain:
What's the thought behind using `pbx` over the current `protons` module in this PR?

I'm looking at performance optimisations and our protobuf serialization is a very hot code path so will be looking at that soon.  I don't want to spend time optimising the wrong module!

vmx:
> What's the thought behind using `pbf` over the current `protons` module in this PR?

The main reason was that I didn't want a dependency on Node.js Buffers but be able to use ArrayBuffers. I also prefer not to maintain forks of something. In addition to that the mapbox folks (who are maintaining pbf) really know what they are doing. And finally I preferred having the schema compiled once and committed rather then doing it on load/runtime.

achingbrain:
If the `protons` -> `pbf` change is worth doing it could be pulled out of this PR and done separately, right?  No need to break the public API for an internal serialisation library change.

Obviously if this module starts emitting `ArrayBuffer`s instead of Node `Buffer`s it's going to upset things, but we can `Buffer.from` before outputting things.  According to the [docs](https://nodejs.org/api/buffer.html#buffer_class_method_buffer_from_arraybuffer_byteoffset_length) it'll create a view on the `ArrayBuffer` instead of copying it.

vmx:
Superseded by https://github.com/ipld/js-ipld-dag-pb/pull/127.

Issue #107, opened 2018-12-10
avoid high buffer allocations in linkSort


mcollina:
In `linkSort` https://github.com/ipld/js-ipld-dag-pb/blob/39cfef1757f733649afcd00bfd02b64270145b4b/src/dag-node/util.js#L27-L32, 2 new buffers are allocated for each comparison. When sorting a big array (100+ entries, which seems common), this is going to create very high gc pressure.
Ideally a sort implementation should not allocate any new objects at all. 

See #104 for a bigger discussion on the sorting logic.

achingbrain:
`link.name` should be a string so I wonder why it doesn't just use `String.prototype.localeCompare()`

@vmx any ideas?

vmx:
It could be related to the Go implementation. We should use the same algorithm in all implementations.

Issue #104, opened 2018-12-05
Stable sorting


mcollina:
the `stable` module keeps coming up in my flamegraphs, and I was wondering if it's truly needed here.
It was part of the initial implementation https://github.com/ipld/js-ipld-dag-pb/commit/cc914b162f85bb97b839bfd3a32cf174a07f6e8a. @daviddias maybe you can chime in on this one?

If it's truly needed, since Node v11 `Array.sort` is stable, so we should prefer that instead.

vmx:
FYI: I currently heavily rework this module. So I'd expect the performance characteristics being quite different (hopefully soon).

But that the sorting is a bottle-neck is certainly good to know.

alanshaw:
It doesn't look to me as though it's needed, since the only [sorting comparison function comes from this module](https://github.com/ipld/js-ipld-dag-pb/blob/39cfef1757f733649afcd00bfd02b64270145b4b/src/dag-node/create.js#L8) and we [`map` any input to a brand new array immediately before calling `stable` sort](https://github.com/ipld/js-ipld-dag-pb/blob/39cfef1757f733649afcd00bfd02b64270145b4b/src/dag-node/create.js#L28-L31).

@mcollina what do you think? I'd appreciate your expertise here.

mcollina:
+1 for removal, but I'm not familiar with the internals enough to understand what side effect it may have.

@daviddias ?

mcollina:
This seems a good approach to avoid slowdown for big files, but not for directories with a ton of small files.

I’d also do a quick check if adding things in order might be better than sorting after the fact.

How big is normally that array?

achingbrain:
For UnixFS directories it can be up to 1000 before it's converted to a sharded directory, then it'll be max 255.

For UnixFS files created by the importer it's max 174.

I don't think there's anything to stop a user creating arbitrarily large lists of links, though I could be wrong.  Might be something to add.

mcollina:
One last question: not being compatible with the Go implementation on this bit would cause any problems to users? If two links are not in the same order, will there be any issues?

achingbrain:
I believe we want to maintain compatibility with go here.  All implementations should return the same hash for a given data structure.

PR #110, opened 2018-12-14
perf: memoize the Name buffer in DAGLink to avoid unneeded allocations

mcollina:
Allocating two buffers for every comparison was quite expensive.
This commit replaces Buffer.compare with String.prototype.localeCompare.

Fixes https://github.com/ipld/js-ipld-dag-pb/issues/107
Fixes: #104


mcollina:
Added the comment.

PR #114, opened 2019-01-26
Import async/setImmediate to avoid webpack errors

PiusNyakoojo:


vmx:
Thanks @PiusNyakoojo for your contribution. With the release of 0.16 this is no longer needed as we don't use callbacks, but a promised based API.

PR #127, opened 2019-04-21
feat: new IPLD Format API

vmx:
BREAKING CHANGE: The API is now async/await based

There are numerous changes, the most significant one is that the API
is no longer callback based, but it using async/await.

The properties of DAGNode and DAGLink are now in sync with the paths
that are used for resolving. This means that e.g. `name` is now called
`Name` and `size` is `Tsize`.

All return values from `resolve()` now conform to the [IPLD Data Model],
this means that e.g. links are no longer represented as
`{'/': "baseecodedcid"}`, but as [CID] instances instead.

For the full new API please see the [IPLD Formats spec].

[IPLD Data Model]: https://github.com/ipld/specs/blob/master/IPLD-Data-Model-v1.md
[CID]: https://github.com/multiformats/js-cid/
[IPLD Formats spec]: https://github.com/ipld/interface-ipld-format


src/dag-node/addLink.js
```diff
@@ -8,40 +8,30 @@ const DAGLink = require('../dag-link')
 const DAGNode = require('./index')
 const create = require('./create')
 
-function asDAGLink (link, callback) {
+const asDAGLink = async (link) => {
```
rvagg: This is totally a matter of preference but I'd have just add the `async` to the existing function. I tend to reserve arrow functions for when I need an anonymous function and/or closure. I'm not thrilled about the disappearance of `function` across many code bases, I believe it reduces clarity a little bit for the sake of a few characters and something that feels new. Up to you of course!

src/dag-link/index.js
```diff
@@ -16,25 +17,29 @@ class DAGLink {
     this._nameBuf = null
     this._size = size
     this._cid = new CID(cid)
+
+    // Make sure we have a nice public API that can be used by an IPLD resolver
+    visibility.hidePrivateFields(this)
+    visibility.addEnumerableGetters(this, ['Hash', 'Name', 'Tsize'])
```
rvagg: why are we going with capitalised names? is there something special about dag-pb that warrants this, otherwise it seems like it's straying into non-idiomatic-land.

src/dag-node/addLink.js
rvagg: `async` can be removed from this function since `await` isn't used, the call to `toDagLink` will return a `Promise` which will be returned and treated appropriately by the calling function which uses `await` to handle it (or the non-`Promise` returns)

src/dag-node/create.js
```diff
@@ -9,36 +9,25 @@ const linkSort = dagNodeUtil.linkSort
 const DAGNode = require('./index.js')
 const DAGLink = require('../dag-link')
 
-function create (data, links, callback) {
-  if (typeof data === 'function') {
-    callback = data
-    data = undefined
-  } else if (typeof data === 'string') {
+const create = async (data, links = []) => {
```
rvagg: `async` can be removed from this because there's only one `await`, which itself is unnecessary and should be removed

src/dag-node/create.js
```diff
@@ -9,36 +9,25 @@ const linkSort = dagNodeUtil.linkSort
...
 
-  serialize({
-    data, links
-  }, (err, buffer) => {
-    if (err) {
-      return callback(err)
-    }
-
-    return callback(null, new DAGNode(data, links, buffer.length))
+  const serialized = await serialize({
```
rvagg: `serialize()` doesn't return a `Promise` at all so `await` can be removed from this

src/resolver.js
```diff
@@ -1,164 +1,70 @@
...
-
-        const value = values[split[0]]
-
-        if (value) {
-          return cb(null, {
-            value: { '/': value.cid },
-            remainderPath: split.slice(1).join('/')
-          })
-        }
+exports.resolve = async (binaryBlob, path) => {
```
rvagg: `async` can be removed from this, it'll just make an unnecessary `Promise`

src/resolver.js
```diff
@@ -1,164 +1,70 @@
...
+    remainderPath: ''
+  }
+}
 
-    callback(null, paths)
-  })
+const traverse = function * (node, path) {
+  // Traverse only objects and arrays
+  if (Buffer.isBuffer(node) || CID.isCID(node) || typeof node === 'string' ||
+      node === null) {
```
rvagg: `==`

test/dag-node-test.js
```diff
@@ -32,459 +29,261 @@ module.exports = (repo) => {
   const bs = new BlockService(repo)
 
   describe('DAGNode', () => {
-    it('create a node', (done) => {
-      expect(7).checks(done)
+    it('create a node', async () => {
```
rvagg: `async` can be removed

test/dag-node-test.js
```diff
@@ -32,459 +29,261 @@ module.exports = (repo) => {
...
       const data = Buffer.from('some data')
 
-      DAGNode.create(data, (err, node) => {
-        expect(err).to.not.exist.mark()
-        expect(node.data.length).to.be.above(0).mark()
-        expect(Buffer.isBuffer(node.data)).to.be.true.mark()
-        expect(node.size).to.be.above(0).mark()
+      const node = await DAGNode.create(data)
```
rvagg: `await` can be removed

test/dag-node-test.js
```diff
@@ -32,459 +29,261 @@ module.exports = (repo) => {
...
-        })
-      })
+      const serialized = dagPB.util.serialize(node)
+      const deserialized = dagPB.util.deserialize(serialized)
+      expect(node.Data).to.eql(deserialized.Data)
     })
 
-    it('create a node with string data', (done) => {
-      expect(7).checks(done)
+    it('create a node with string data', async () => {
```
rvagg: `async` can be removed

test/dag-node-test.js
```diff
@@ -32,459 +29,261 @@ module.exports = (repo) => {
...
       const data = 'some data'
 
-      DAGNode.create(data, (err, node) => {
-        expect(err).to.not.exist.mark()
-        expect(node.data.length).to.be.above(0).mark()
-        expect(Buffer.isBuffer(node.data)).to.be.true.mark()
-        expect(node.size).to.be.above(0).mark()
+      const node = await DAGNode.create(data)
```
rvagg: `await` can be removed

test/dag-node-test.js
```diff
@@ -32,459 +29,261 @@ module.exports = (repo) => {
...
-            expect(node.data).to.eql(deserialized.data).mark()
-          })
-        })
-      })
+      const deserialized = dagPB.util.deserialize(serialized)
+      expect(node.Data).to.eql(deserialized.Data)
     })
 
-    it('create a node with links', (done) => {
+    it('create a node with links', async () => {
```
rvagg: `async` can be removed

test/dag-node-test.js
```diff
@@ -32,459 +29,261 @@ module.exports = (repo) => {
...
-          'some link',
-          'some other link'
-        ])
-        done()
+      const node1 = await DAGNode.create(someData, l1)
+      const l2 = l1.map((l) => {
+        return new DAGLink(l.Name, l.Tsize, l.Hash)
       })
+
+      const node2 = await DAGNode.create(someData, l2)
```
rvagg: `await` can be removed

test/dag-node-test.js
```diff
@@ -32,459 +29,261 @@ module.exports = (repo) => {
...
+      // check sorting
+      expect(node1.Links.map((l) => l.Name)).to.be.eql([
+        'some link',
+        'some other link'
+      ])
     })
 
-    it('create with empty link name', (done) => {
-      DAGNode.create(Buffer.from('hello'), [
+    it('create with empty link name', async () => {
```
rvagg: `async` can be removed

test/dag-node-test.js
```diff
@@ -32,459 +29,261 @@ module.exports = (repo) => {
...
+      const node = await DAGNode.create(Buffer.from('hello'), [
```
rvagg: `await` can be removed

test/dag-node-test.js
```diff
@@ -32,459 +29,261 @@ module.exports = (repo) => {
...
-              Object.getOwnPropertyNames(node).forEach(key => {
-                expect(node[key]).to.deep.equal(deserialized[key])
-              })
-
-              cb()
-            }
-          ], cb)
-        }
-      ], done)
+    it('create with undefined link name', async () => {
```
rvagg: `async` can be removed and `await` from next line

test/dag-node-test.js
```diff
@@ -32,459 +29,261 @@ module.exports = (repo) => {
...
+      expect(node.Links[0].Name).to.be.eql('')
+      const serialized = dagPB.util.serialize(node)
+      const deserialized = dagPB.util.deserialize(serialized)
+      for (const key of Object.keys(node)) {
+        expect(node[key]).to.deep.equal(deserialized[key])
+      }
     })
 
-    it('create an empty node', (done) => {
+    it('create an empty node', async () => {
```
rvagg: `async` can be removed

test/dag-node-test.js
```diff
@@ -32,459 +29,261 @@ module.exports = (repo) => {
...
-        dagPB.util.serialize(node, (err, serialized) => {
-          expect(err).to.not.exist.mark()
-
-          dagPB.util.deserialize(serialized, (err, deserialized) => {
-            expect(err).to.not.exist.mark()
-            expect(node.data).to.eql(deserialized.data).mark()
-          })
-        })
-      })
+      const node = await DAGNode.create(Buffer.alloc(0))
```
rvagg: `await` can be removed

test/dag-node-test.js
```diff
@@ -32,459 +29,261 @@ module.exports = (repo) => {
...
-      DAGNode.create({}, (err, node) => {
-        expect(err).to.exist()
-        expect(node).to.not.exist()
-        DAGNode.create([], (err, node) => {
-          expect(err).to.exist()
-          expect(node).to.not.exist()
-          done()
-        })
-      })
+    it('fail to create a node with other data types', async () => {
```
rvagg: `async` can be removed

test/dag-node-test.js
```diff
@@ -32,459 +29,261 @@ module.exports = (repo) => {
...
+      await expect(DAGNode.create({})).to.be.rejectedWith(
```
rvagg: these should be converted to `throws` tests

test/dag-node-test.js
```diff
@@ -32,459 +29,261 @@ module.exports = (repo) => {
...
-            expect(err).to.not.exist()
-            expect(node1b.links.length).to.equal(1)
-            expect(node1b.links[0].size)
-              .to.eql(node2.size)
-            expect(node1b.links[0].name).to.be.eql('')
-            cb()
-          })
-        }
-      ], done)
+    it('addLink by DAGNode', async () => {
```
rvagg: `async` can be removed and the next 3 `await`s can be removed

test/dag-node-test.js
```diff
@@ -32,459 +29,261 @@ module.exports = (repo) => {
...
-            expect(node1b.links.length).to.equal(1)
-            expect(node1b.links[0].size)
-              .to.eql(node2.size)
-            expect(node1b.links[0].name).to.be.eql('')
-            cb()
-          })
-        }
-      ], done)
+    it('addLink by DAGLink', async () => {
+      const node1a = await DAGNode.create(Buffer.from('1'))
```
rvagg: this and the next `await` can be removed

test/dag-node-test.js
```diff
@@ -32,459 +29,261 @@ module.exports = (repo) => {
...
+      expect(node1b.Links.length).to.equal(1)
+      expect(node1b.Links[0].Tsize).to.eql(node2.size)
+      expect(node1b.Links[0].Name).to.be.eql('')
     })
 
-    it('addLink by object', (done) => {
-      let node1
-      let node2
+    it('addLink by object', async () => {
+      const node1a = await DAGNode.create(Buffer.from('1'))
```
rvagg: this and the next `await` can be removed

test/dag-node-test.js
```diff
@@ -32,459 +29,261 @@ module.exports = (repo) => {
...
-            expect(node1b.links.length).to.equal(1)
-            expect(node1b.links[0].size)
-              .to.eql(node2.size)
-            expect(node1b.links[0].name).to.be.eql('')
-            cb()
-          })
-        }
-      ], done)
+    it('addLink by name', async () => {
+      const node1a = await DAGNode.create(Buffer.from('1'))
```
rvagg: this and the next `await` can be removed

test/dag-node-test.js
```diff
@@ -32,459 +29,261 @@ module.exports = (repo) => {
...
+      expect(node1b.Links[0].Name).to.eql('banana')
+      expect(Object.keys(node1b)).to.include('banana')
     })
 
-    it('addLink - add several links', (done) => {
-      let node1a
-      let node1b
-      let node1c
+    it('addLink - add several links', async () => {
+      const node1a = await DAGNode.create(Buffer.from('1'))
```
rvagg: `await` can be removed

test/dag-node-test.js
```diff
@@ -32,459 +29,261 @@ module.exports = (repo) => {
...
+      expect(node1b.Links[0].Tsize).to.eql(node2.size)
+      expect(node1b.Links[0].Name).to.eql('banana')
+      expect(Object.keys(node1b)).to.include('banana')
     })
 
-    it('addLink - add several links', (done) => {
-      let node1a
-      let node1b
-      let node1c
+    it('addLink - add several links', async () => {
```
rvagg: `async` can be removed

test/dag-node-test.js
```diff
@@ -32,459 +29,261 @@ module.exports = (repo) => {
...
-        },
-        (cb) => {
-          expect(node1a.links.length).to.equal(0)
-          expect(node1b.links.length).to.equal(1)
-          expect(node1c.links.length).to.equal(2)
-          cb()
-        }
-      ], done)
-    })
+      const node2 = await DAGNode.create(Buffer.from('2'))
```
rvagg: `await` can be removed

test/dag-node-test.js
```diff
@@ -32,459 +29,261 @@ module.exports = (repo) => {
...
+      const node1b = await DAGNode.addLink(node1a, node2)
+      expect(node1b.Links.length).to.equal(1)
 
-    it('rmLink by name', (done) => {
-      let node1a
-      let node1b
-      let node2
+      const node3 = await DAGNode.create(Buffer.from('3'))
```
rvagg: `await` can be removed

test/dag-node-test.js
```diff
@@ -32,459 +29,261 @@ module.exports = (repo) => {
...
-        (cb) => {
-          DAGNode.rmLink(node1b, 'banana', (err, node) => {
-            expect(err).to.not.exist()
-            expect(node1a.toJSON()).to.eql(node.toJSON())
-            cb()
-          })
-        }
-      ], done)
+    it('rmLink by name', async () => {
+      const node1a = await DAGNode.create(Buffer.from('1'))
```
rvagg: `await` can be removed

test/dag-node-test.js
```diff
@@ -32,459 +29,261 @@ module.exports = (repo) => {
...
+      expect(node1a.Links.length).to.eql(0)
+      const withoutLink = node1a.toJSON()
+
+      const node2 = await DAGNode.create(Buffer.from('2'))
```
rvagg: `await` can be removed

src/dag-node/rmLink.js
```diff
@@ -6,19 +6,19 @@ const cloneData = dagNodeUtil.cloneData
 const create = require('./create')
 const CID = require('cids')
 
-function rmLink (dagNode, nameOrCid, callback) {
+const rmLink = async (dagNode, nameOrCid) => {
```
rvagg: `async` can be removed from this, there's nothing async in the call path

test/dag-node-test.js
```diff
@@ -32,459 +29,261 @@ module.exports = (repo) => {
...
+      const link = await toDAGLink(node2, { name: 'banana' })
+
+      const node1b = await DAGNode.addLink(node1a, link)
+      expect(node1b.Links.length).to.eql(1)
+      expect(Object.keys(node1b)).to.include('banana')
+      const node1c = await DAGNode.rmLink(node1b, 'banana')
```
rvagg: `await` an be removed

test/dag-node-test.js
```diff
@@ -32,459 +29,261 @@ module.exports = (repo) => {
...
+      expect(node1c.Links.length).to.eql(0)
+      expect(node1c.toJSON()).to.eql(withoutLink)
     })
 
-    it('rmLink by hash', (done) => {
-      let node1a
-      let node1b
-      let node2
+    it('rmLink by hash', async () => {
+      const node1a = await DAGNode.create(Buffer.from('1'))
```
rvagg: `await` can be removed

test/dag-node-test.js
```diff
@@ -32,459 +29,261 @@ module.exports = (repo) => {
...
+      expect(node1a.Links.length).to.eql(0)
+      const withoutLink = node1a.toJSON()
+
+      const node2 = await DAGNode.create(Buffer.from('2'))
```
rvagg: `await` can be removed

test/dag-node-test.js
```diff
@@ -32,459 +29,261 @@ module.exports = (repo) => {
...
-        (cb) => {
-          DAGNode.rmLink(node1b, node1b.links[0].cid, (err, node) => {
-            expect(err).to.not.exist()
-            expect(node1a.toJSON()).to.eql(node.toJSON())
-            cb()
-          })
-        }
-      ], done)
+    it('get node CID', async () => {
+      const node = await DAGNode.create(Buffer.from('some data'))
```
rvagg: `await` can be removed

test/dag-node-test.js
```diff
@@ -32,459 +29,261 @@ module.exports = (repo) => {
...
-          expect(cid.multihash).to.exist()
-          expect(cid.codec).to.equal('dag-pb')
-          expect(cid.version).to.equal(1)
-          const mh = multihash.decode(cid.multihash)
-          expect(mh.name).to.equal('sha2-512')
-          done()
-        })
-      })
+    it('get node CID with hashAlg', async () => {
+      const node = await DAGNode.create(Buffer.from('some data'))
```
rvagg: `await` can be removed

test/dag-node-test.js
```diff
@@ -32,459 +29,261 @@ module.exports = (repo) => {
...
-            })
-          }),
-          ({ buffer, cid }, cb) => {
-            block = new Block(buffer, cid)
-            bs.put(block, cb)
-          },
-          (cb) => bs.get(block.cid, cb),
-          (retrievedBlock, cb) => {
+    it('marshal a node and store it with block-service', async () => {
+      const node = await DAGNode.create(Buffer.from('some data'))
```
rvagg: `await` can be removed

test/dag-node-test.js
```diff
@@ -32,459 +29,261 @@ module.exports = (repo) => {
...
+      const serialized = dagPB.util.serialize(node)
+      const cid = await dagPB.util.cid(serialized)
+      const block = new Block(Buffer.from(serialized), cid)
+      return new Promise((resolve) => {
```
rvagg: A good rule of thumb is that if you're ever using `new Promise` then you're probably doing something wrong.

Dealing with callback interfaces can make it tricky, but trying to manually wrap them in Promises can lead easily to errors getting lost. But that's where [`util.promisify()`](https://nodejs.org/api/util.html#util_util_promisify_original) comes in handy and it'll translate into the browser too.

Take it or leave it, but here's a patch that addresses this and makes the `BlockService` interface `await`-ready and error safe:

```patch
diff --git a/test/dag-node-test.js b/test/dag-node-test.js
index 5910e28..8fe66fd 100644
--- a/test/dag-node-test.js
+++ b/test/dag-node-test.js
@@ -1,6 +1,7 @@
 /* eslint-env mocha */
 'use strict'
 
+const { promisify } = require('util')
 const chai = require('chai')
 const chaiAsProised = require('chai-as-promised')
 const dirtyChai = require('dirty-chai')
@@ -26,7 +27,11 @@ const testBlockNamedLinks = loadFixture('test/fixtures/test-block-named-links')
 const testBlockUnnamedLinks = loadFixture('test/fixtures/test-block-unnamed-links')
 
 module.exports = (repo) => {
-  const bs = new BlockService(repo)
+  const _bs = new BlockService(repo)
+  const bs = {
+    get: promisify(_bs.get.bind(_bs)),
+    put: promisify(_bs.put.bind(_bs))
+  }
 
   describe('DAGNode', () => {
     it('create a node', async () => {
@@ -252,16 +257,9 @@ module.exports = (repo) => {
       const serialized = dagPB.util.serialize(node)
       const cid = await dagPB.util.cid(serialized)
       const block = new Block(Buffer.from(serialized), cid)
-      return new Promise((resolve) => {
-        bs.put(block, (err) => {
-          expect(err).to.not.exist()
-          bs.get(block.cid, (err, retrievedBlock) => {
-            expect(err).to.not.exist()
-            expect(retrievedBlock).to.eql(block)
-            resolve()
-          })
-        })
-      })
+      await bs.put(block)
+      const retrievedBlock = await bs.get(block.cid)
+      expect(retrievedBlock).to.eql(block)
     })
 
     it('deserialize go-ipfs block from ipldResolver', async () => {
```

test/dag-node-test.js
```diff
@@ -32,459 +29,261 @@ module.exports = (repo) => {
...
       }
 
       const cidStr = 'QmYwAPJzv5CZsnA625s3Xf2nemtYgPpHdWEz79ojWnPbdG'
       const cid = new CID(cidStr)
 
-      bs.get(cid, (err, block) => {
-        expect(err).to.not.exist()
-        dagPB.util.deserialize(block.data, (err, node) => {
+      return new Promise((resolve) => {
+        bs.get(cid, (err, block) => {
```
rvagg: if using the `promisify` wrapper, same thing applies here and you can ditch the error.to.not.exists and returns

test/dag-node-test.js
```diff
@@ -564,41 +359,31 @@ module.exports = (repo) => {
...
 
-    it('dagNode.toJSON with empty Node', (done) => {
-      DAGNode.create(Buffer.alloc(0), (err, node) => {
-        expect(err).to.not.exist()
-        expect(node.toJSON().data).to.eql(Buffer.alloc(0))
-        expect(node.toJSON().links).to.eql([])
-        expect(node.toJSON().size).to.exist()
-        done()
-      })
+    it('dagNode.toJSON with empty Node', async () => {
```
rvagg: `async` and the `await` on the next line can be removed

test/dag-node-test.js
```diff
@@ -564,41 +359,31 @@ module.exports = (repo) => {
...
+      const node = await DAGNode.create(Buffer.alloc(0))
+      expect(node.toJSON().data).to.eql(Buffer.alloc(0))
+      expect(node.toJSON().links).to.eql([])
+      expect(node.toJSON().size).to.exist()
     })
 
-    it('dagNode.toJSON with data no links', (done) => {
+    it('dagNode.toJSON with data no links', async () => {
```
rvagg: `async` can be removed

test/dag-node-test.js
```diff
@@ -564,41 +359,31 @@ module.exports = (repo) => {
...
       const data = Buffer.from('La cucaracha')
-      DAGNode.create(data, (err, node) => {
-        expect(err).to.not.exist()
-        expect(node.toJSON().data).to.eql(data)
-        expect(node.toJSON().links).to.eql([])
-        expect(node.toJSON().size).to.exist()
-        done()
-      })
+      const node = await DAGNode.create(data)
```
rvagg: `await` can be removed

test/dag-node-test.js
```diff
@@ -564,41 +359,31 @@ module.exports = (repo) => {
...
+      expect(node.toJSON().data).to.eql(data)
+      expect(node.toJSON().links).to.eql([])
+      expect(node.toJSON().size).to.exist()
     })
 
-    it('add two nameless links to a node', (done) => {
+    it('add two nameless links to a node', async () => {
```
rvagg: `await` can be removed

test/dag-node-test.js
```diff
@@ -611,56 +396,41 @@ module.exports = (repo) => {
...
+      const link1 = new DAGLink(l1.Name, l1.Tsize,
+        Buffer.from(bs58.decode(l1.Hash)))
+      const link2 = new DAGLink(l2.Name, l2.Tsize,
+        Buffer.from(bs58.decode(l2.Hash)))
 
-      DAGNode.create(Buffer.from('hiya'), [link1, link2], (err, node) => {
-        expect(err).to.not.exist()
-        done()
-      })
+      const node = await DAGNode.create(Buffer.from('hiya'), [link1, link2])
```
rvagg: `await` can be removed

test/dag-node-test.js
```diff
@@ -611,56 +396,41 @@ module.exports = (repo) => {
...
     })
 
-    it('toString', (done) => {
-      DAGNode.create(Buffer.from('hello world'), (err, node) => {
-        expect(err).to.not.exist()
-        const expected = 'DAGNode <data: "aGVsbG8gd29ybGQ=", links: 0, size: 13>'
-        expect(node.toString()).to.equal(expected)
-        done()
-      })
+    it('toString', async () => {
```
rvagg: `async` and the `await` on the next line can be removed

test/dag-node-test.js
```diff
@@ -611,56 +396,41 @@ module.exports = (repo) => {
...
+      const node = await DAGNode.create(Buffer.from('hello world'))
+      const expected = 'DAGNode <data: "aGVsbG8gd29ybGQ=", links: 0, size: 13>'
+      expect(node.toString()).to.equal(expected)
     })
 
-    it('deserializing a node and an object should yield the same result', (done) => {
-      expect(10).checks(done)
+    it('deserializing a node and an object should yield the same result', async () => {
```
rvagg: `async` can be removed

test/dag-node-test.js
```diff
@@ -611,56 +396,41 @@ module.exports = (repo) => {
...
-                expect(err).to.not.exist.mark()
-                expect(deserialized.toJSON()).to.deep.equal(deserializedObject.toJSON()).mark()
-                done()
-              })
-            })
-          })
-        })
-      })
-    }).timeout(6000)
+      const node = await DAGNode.create(obj.Data, obj.Links)
```
rvagg: `await` can be removed

test/resolver.spec.js
```diff
@@ -22,282 +19,209 @@ describe('IPLD Format resolver (local)', () => {
...
-    })
-  })
-
-  it('multicodec is dag-pb', () => {
-    expect(resolver.multicodec).to.equal('dag-pb')
-  })
-
-  it('defaultHashAlg is sha2-256', () => {
-    expect(resolver.defaultHashAlg).to.equal('sha2-256')
+  const create = async (data, links) => {
```
rvagg: I should probably have stopped earlier re `async` and `await`, will put thoughts about async/await in the main comments

test/resolver.spec.js
```diff
@@ -22,282 +19,209 @@ describe('IPLD Format resolver (local)', () => {
...
-    it('resolver.tree', (done) => {
-      resolver.tree(emptyNodeBlob, (err, paths) => {
-        expect(err).to.not.exist()
-        expect(paths).to.eql([
-          'Links',
-          'Data'
-        ])
-        done()
-      })
+    it('resolver.tree', async () => {
```
rvagg: no `await` in here, `async` can be removed

test/resolver.spec.js
```diff
@@ -22,282 +19,209 @@ describe('IPLD Format resolver (local)', () => {
...
-          'Links/0/Tsize',
-          'Links/0/Hash',
-          'Links/1/Name',
-          'Links/1/Tsize',
-          'Links/1/Hash',
-          'Data'
-        ])
-        done()
-      })
+    it('resolver.tree', async () => {
```
rvagg: no `await` in here, `async` can be removed

rvagg:
Sorry, there's a bunch of noise in there.

First, the Title cased properties aren't great. Title case in idiomatic JavaScript should be reserved for class / prototype / global object names, elsewhere it's just out of place.

Next, the async thing is really interesting. This is the only truly set of async functions in the whole lot:

`addLink()`
-> `asDagLink()`
-> `toDagLink()`
-> `cid()`
-> `multihashing()`

And it's only in `addLink()` as sugar, arguably.

One nice thing about `await` is that it's agnostic about whether it's really waiting on a `Promise` or not. So you can export functions that aren't `async` and use them as if they are. So the ipld-dag formats interface may assume it, but the formats themselves don't necessarily need to implement it that way. As long as the user uses `await` they can be `async` or bare. When using a maybe-`async` interface, you can squeeze out extra performance by inspecting a return value to see if it's a `Promise` or not and either resolve/`await` it or use the bare value if it's not (this is the discussion we've been having about the new stack interfaces too). IMO that gets a bit messy but because `await` will insert an extra `Promise` (or two, or three, depending on the VM), it can be worthwhile in hot code.

So, in the case of ipld-dag-pb, the only thing we're doing that is async is some `multihashing-async` work. So maybe we could consider adjusting the interface slightly stop that getting in the way of `addLink()` by refusing to add a `DAGNode` as a link and forcing the upstream user to do the conversion themselves prior to calling `addLink()`. Then any user code that doesn't need to do such a conversion will get a fully-sync interface. This would be as simple as something like exposing `async toDAGLink()` on `DAGNode` and then callers just need to replace `await DAGNode.addLink(node2, node1)` with `DAGNode.addLink(node2, await node1.toDAGLink())` and then all other strictly necessary `await` calls go away leaving them optional.

It'd be pretty neat to ditch all of these pointless `setInterval()` calls unnecessary `callback` arguments with an interface that has a single method that returns a `Promise`. That'd have a pretty massive perf impact if ipld-dag-pb is being used so heavily.

rvagg:
OK, I see where the Titlecase is coming from, it's in the PB data model https://github.com/ipld/specs/pull/114 (and util.js). I still don't think that should leak out to the JS interface though, it should be the job of the interface to translate a data layer into a language-idiomatic presentation of it.

src/dag-node/addLink.js
vmx: Even if the `async` is not needed I like it as an indicator for "this thing returns a Promise". If I remove that, the caller has no indication that a Promise is returned. Am I missing something?

vmx: Sorry I missed that this function isn't async at all.

vmx: I still don't get it. I've created this minimal example, that should resemble the calls mentioned here. If I don't have an `async` at `asDagLink()` it will still return a promise:

```js
class Something {
  constructor (value) {
    this.value = value
  }
}

const cid = async () => {
  return 'test'
}

const toDAGLink = async () => {
  const result = await cid()
  return new Something(result)
}

const asDAGLink = () => {
  return toDAGLink()
}

const output = asDAGLink()
console.log(output) // Promise { <pending> }
```

src/resolver.js
vmx: It's for consistency with other formats. I'd like to keep it for now until it turns out to be an issue.

test/dag-node-test.js
vmx: Wow thanks a lot for the code. I've used the `promisify-es6` module which is also used in js-ipfs/js-ipld which leads to a smaller browser bundle compared to the Node.js polyfill.

vmx:
I've opened an issue for further `addLink()` discussions: https://github.com/ipld/js-ipld-dag-pb/issues/128. I won't like to postpone the merge of this PR even more.

vmx:
I also don't like those Camelcase property names. But it makes sense in regards how current IPLD Formats are envisioned. If they serialize/deserialize synchronously then `resolve()` and `tree()` can just be generic functions. Also I consider `dag-pb` legacy and it should be only consumed by IPFS. I think it's OK to have a less JSy interface there.

vmx:
@rvagg For future reference, feel free to mention things like "`await` not needed here" with a comment that it is throughout the whole file, you don't need to repeat it.

src/dag-node/addLink.js
rvagg: I think in this case it comes down to a matter of taste. I can't say for sure that adding `async` will add another `Promise` on top of the one that's returned, I'd hope that VMs optimise that away.
I'm personally happy leaving off `async` from a function when it doesn't contain an `await` and letting the documentation do the work of saying that it returns a `Promise`. Unclear return types are one of the many costs of an untyped language (and `async` only tells you that you can `await` it, not what you're `await`ing, so even this is limited), along with argument types, and we're just used to having to look at metadata (documentation usually) to tell us what an API requires and will return.
And again, `@async` and `@returns Promise` in JSDoc are good tools and can live with your code.

rvagg: Actually ... this function also does a bare return in some cases `return new DAGLink...`, which is fine if you're using `await` but might cause problems for users who expect a `Promise`. So yeah, consistency imposed by `async` is probably best. It'd be like if your example was:

```js
const asDAGLink = () => {
  return Math.random() > 0.5 ? false : toDAGLink()
}
```

Which would be a mess of an API.

src/resolver.js
rvagg: I get the consistency argument but really think the format API needs to be "MaybePromise" rather than "Promise", or just make `resolve()` completely synchronous. None of the formats do anything async within `resolve()`, they're all the same pattern as this one (other than raw which is even simpler):

* https://github.com/ipld/js-ipld-dag-cbor/pull/107/files#diff-fe783c782beb75b01586f9afb052da1bR21
* https://github.com/ipld/js-ipld-zcash/pull/39/files#diff-fe783c782beb75b01586f9afb052da1bR21
* https://github.com/ipld/js-ipld-bitcoin/pull/48/files#diff-fe783c782beb75b01586f9afb052da1bR21
* https://github.com/ipld/js-ipld-git/pull/51/files#diff-fe783c782beb75b01586f9afb052da1bR21
* https://github.com/ipld/js-ipld-raw/pull/32/files#diff-1fdf421c05c1140f6d71444ea2b27638R22

So the `async` here is theoretical at best. I like that it might be accommodating a potential of encryption or something else novel, but so far it doesn't, and that could be accomplished by stating in the format specification that it "may be a Promise" so users may want to opt for `await` if they want to be maximally compatible. I suspect most uses of the current code are sticking to specific implementations though and that flexibility can be left to the next generation stack.

As I understand, Matteo's performance analysis identified as the faux-async with `setImmediate()` everywhere was the biggest cost for js-ipfs. Replacing them with `async` doesn't buy much at all, it's just different sugar for deferring something that isn't actually asynchronous.

rvagg:
My remaining hangup is with `resolve()` being `async`, expanded on in https://github.com/ipld/js-ipld-dag-pb/pull/127/files#r278797728. I first question what the rationale for making it `async` in the format spec is if we don't have a current format that needs it. Then, if we have a theoretical need for it (encryption, maybe), I advocate for an optional `async` and documenting that users should use `await` or do `Promise` inspection if they are not depending on a specific format but need to be generic. `Promise`s have a big perf cost in cases where they're not needed and this lowest level of ipfs is going to have a big impact. If the current unixfs implementation is tied to dag-pb then it could be refactored to entirely do away with the faux-async in every case other than CID generation and it'd have huge performance gains. By forcing an `async` there's no option for downstream consumers to avoid the cost.

rvagg:
`clone()` in src/dag-node/clone.js still takes a callback and calls `create()` with a callback.

And README needs updating, is that a separate aegir step?

src/resolver.js
vmx: I'll make `resolve()` sync for the same reason that I made `serialize()/deserialize()` sync. Thanks for this outside perspective. I'm currently too deep into this without seeing the bigger picture.

PR #138, opened 2019-05-20
Remove cloning

vmx:
BREAKING CHANGE: `DAGNode.clone()` is removed from public API without any replacement.

Also the API for `rmLink()` and `addLink()` changed. They no longer
return a new node, but just remove/add the links to/from the current
node.

Prior to this change:

    const lessLinks = DAGNode.rmLink(node1, 'Link1')
    node1 = lessLinks
    const moreLinks = await DAGNode.addLink(node2, link)
    node2 = moreLinks

Now:

    DAGNode.rmLink(node, 'Link1')
    await DAGNode.addLink(node2, link)


achingbrain:
> Also the API for rmLink() and addLink() changed. They no longer return a new node, but just remove/add the links to/from the current node.

If they don't return new nodes any more, then `DAGLink`s are no longer immutable, so what's the point of having these methods?  Much more straightforward to just use the Array methods:

```javascript
node1.Links = node1.Links.filter(link => link.name !== 'Link1')

node2.Links.push(link)
```

vmx:
@achingbrain The point is the links need to be sorted. Of course we could sort just right before serialization, but I liked the idea to have things correctly sorted even before that.

vmx:
@achingbrain Oh, and another point. If you add a link, there will also be a property on the object created with the name of the link.

achingbrain:
> we could sort just right before serialization

This is what it used to do and it worked well enough.

> there will also be a property on the object created with the name of the link.

I'm not convinced of the utility of this feature.  I guess the idea is you can navigate through `DAGNode` objects similar to how you'd navigate through `DAGCBOR` ones, eg:  `QMfoo/bar.txt` would translate to `node['bar.txt']`, but a few things spring to mind:

1. There might be a link called 'Data', 'Links' or 'size' present which would conflict.  Looks like [they won't get added](https://github.com/ipld/js-ipld-dag-pb/blob/master/src/dag-node/addNamedLink.js#L14-L17) which means `there will also be a property on the object created` is not true - instead `there might also be a property on the object created` is more accurate.  It's an edge case, but it's still one that needs to be handled, and that complexity is passed to the calling code.
2. Assuming the `unixfs` usecase, the `DAGNode` in question might be a sub-shard of a HAMT shard, in which case there's not a 1-to-1 mapping between path segments and node properties so we have to do a whole bunch of extra processing anyway - accessing `node.F0` at the end of that doesn't save us much.
3.  Finally you can still mutate the contents of the `.Links` array so we should stop kidding ourselves that these data structures are immutable.  Consider:

```javascript
const node = DAGNode.create(Buffer.from('some data'), [{
  Hash: 'QmUxD5gZfKzm8UN4WaguAMAZjw2TzZ2ZUmcqm2qXPtais7',
  TSize: 5,
  Name: 'foo'
}])

node._links[0] = new DAGLink('bar', 5, 'QmUxD5gZfKzm8UN4WaguAMAZjw2TzZ2ZUmcqm2qXPtais7')

console.info(node.foo.Name) // 'bar'
```

You know someone, somewhere will find a way to do this.

achingbrain:
More fun times:

```javascript
const node = DAGNode.create(Buffer.from('some data'), [{
  Hash: 'QmUxD5gZfKzm8UN4WaguAMAZjw2TzZ2ZUmcqm2qXPtais7',
  TSize: 5,
  Name: '_links'
}])

 console.info(node._links)
```

```
    RangeError: Maximum call stack size exceeded
      at ClassIsWrapper.get [as _links] (src/dag-node/addNamedLink.js:21:10)
      at ClassIsWrapper.get [as _links] (src/dag-node/addNamedLink.js:21:23)
      at ClassIsWrapper.get [as _links] (src/dag-node/addNamedLink.js:21:23)
      at ClassIsWrapper.get [as _links] (src/dag-node/addNamedLink.js:21:23)
      at ClassIsWrapper.get [as _links] (src/dag-node/addNamedLink.js:21:23)
      at ClassIsWrapper.get [as _links] (src/dag-node/addNamedLink.js:21:23)
      at ClassIsWrapper.get [as _links] (src/dag-node/addNamedLink.js:21:23)
      at ClassIsWrapper.get [as _links] (src/dag-node/addNamedLink.js:21:23)
      at ClassIsWrapper.get [as _links] (src/dag-node/addNamedLink.js:21:23)
      at ClassIsWrapper.get [as _links] (src/dag-node/addNamedLink.js:21:23)
```

achingbrain:
FWIW the stack overflow is caused because `Object.keys` invoked in `addNamedLink` only returns enumerable properties and the `DAGNode` constructor makes `_` prefixed properties non-enumerable.

It'd be better if these things were just plain JS objects with no funny business.

vmx:
> > we could sort just right before serialization
> 
> This is what it used to do and it worked well enough.

Not really. In the current code `addLink()` always returns a new node, that is run through `create()` which [sorts the links](https://github.com/ipld/js-ipld-dag-pb/blob/d02c65918e537380ad2844b2573283def30098b4/src/dag-node/create.js#L23). So the links within the returned object were always sorted.

> > there will also be a property on the object created with the name of the link.
> 
> I'm not convinced of the utility of this feature. I guess the idea is you can navigate through `DAGNode` objects similar to how you'd navigate through `DAGCBOR` ones, eg: `QMfoo/bar.txt` would translate to `node['bar.txt']`, but a few things spring to mind:

Yes that was the plan. The idea was that whatever you can `resolve()` can also be directly accessed. But as we probably don't really need that feature in dag-pb, I'll just remove it and hard-code everything. That's kind of sad as the whole renaming of properties wouldn't have been needed. Sorry for that.

>     1. There might be a link called 'Data', 'Links' or 'size' present which would conflict.  Looks like [they won't get added](https://github.com/ipld/js-ipld-dag-pb/blob/master/src/dag-node/addNamedLink.js#L14-L17) which means `there will also be a property on the object created` is not true - instead `there might also be a property on the object created` is more accurate.  It's an edge case, but it's still one that needs to be handled, and that complexity is passed to the calling code.

That's not really an issue. The idea is that there's a property for every link you can resolve to. And it was already broken in the way you describe above.

>     3. Finally you can still mutate the contents of the `.Links` array so we should stop kidding ourselves that these data structures are immutable.

I don't think of dag-pb as being immutable.

vmx:
https://github.com/ipld/js-ipld-dag-pb/pull/140 is for the issue about named links mentioned here.

achingbrain:
> I don't think of dag-pb as being immutable.

Then can [this sort of thing](https://github.com/ipld/js-ipld-dag-pb/blob/master/src/dag-node/index.js#L63) be removed?

achingbrain:
> That's kind of sad as the whole renaming of properties wouldn't have been needed. Sorry for that.

Can we change them back before we release `js-ipfs` and we're stuck with them (for a while).

/ducks

vmx:
> Can we change them back before we release `js-ipfs` and we're stuck with them (for a while).

#140 needed less changes than I thought. I like the idea of IPLD being uniform. This means that the enumerable properties are also the ones than can be resolved. This (now) doesn't hold true for named links, but they are pretty special anyway. At least (if we keep the current names) it holds true for those.

vmx:
> > I don't think of dag-pb as being immutable.
> 
> Then can [this sort of thing](https://github.com/ipld/js-ipld-dag-pb/blob/master/src/dag-node/index.js#L63) be removed?

I think for `Data` it can be removed. But I'd still keep it for `Links` as I find setters with side effects (if the side-effect is re-ordering the links) a bit strange.

achingbrain:
You could have both `.data` and `.Data` being dynamic getters/setters that access a non-enumerable `._data` property?  That way you can have uniformity for IPLD and we can have idiomatic js property names that don't expose serialisation concerns to the application layer.

> I think for Data it can be removed. But I'd still keep it for Links

I think you should either have everything mutable or everything immutable, not both.  Inconsistent APIs are the worst.  My preference would be for mutable, because JavaScript.

vmx:
> You could have both `.data` and `.Data` being dynamic getters/setters that access a non-enumerable `._data` property?

I thought about that as well. I still like the idea of having the property with the same name as the path you resolve to, or consistency reason. We can do both, but I personally prefer APIs where there's only one way of doing things. One one hand not breaking APIs might be a case to do both, on the other hand I'd e.g. also like to get rid of DAGLink (but that's another story).

> > I think for Data it can be removed. But I'd still keep it for Links
> 
> I think you should either have everything mutable or everything immutable, not both. Inconsistent APIs are the worst. My preference would be for mutable, because JavaScript.

So you would give up on the sorting for links? So when you add a link via `push()` it would have a different index when you serialize and deserialize the node. I also dislike inconsistent APIs, but the links are sadly a special case.

achingbrain:
> when you add a link via push() it would have a different index when you serialize and deserialize the node.

It sounds weird, I agree, but off the top of my head I can't think of a case where we are creating DAGNodes and then immediately examining them so I don't know how big of an issue it is.

We're either importing files (e.g. serializing) or exporting them (e.g. deserializing).  In the first case it's probably better to defer the sorting until serialization for performance (disclaimer: I haven't measured this).

vmx:
There is no cloning anymore since the merge of https://github.com/ipld/js-ipld-dag-pb/pull/141.

PR #139, opened 2019-05-20
fix: named links should return the CID

vmx:
Prior to this change, named links didn't return the CID they are
pointing to, but the DAGLink. This made proper named links traversal
as it is done by UnixFSv1 impossible.

Thanks @alanshaw for finding and providing a solution to this bug.


vmx:
@achingbrain This might break some of UnixFS, but this is the correct behaviour.

PR #140, opened 2019-05-21
fix: put named links into its own property

vmx:
BREAKING CHANGE: Named links are no longer accessible directly on the object

Named links can still be used for resolving, but those links are no
longer accessible directly on the object.

### DIFF
--- a/src/resolver.js
+++ b/src/resolver.js
@@ -25,10 +25,16 @@ exports.resolve = (binaryBlob, path) => {
   while (parts.length) {
     const key = parts.shift()
     if (node[key] === undefined) {
-      throw new Error(`Object has no property '${key}'`)
+      // Check the named links
+      if (node._namedLinks[key] === undefined) {
+        throw new Error(`Object has no property '${key}'`)
+      } else {
+        node = node._namedLinks[key]
+      }
+    } else {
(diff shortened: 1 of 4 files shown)


src/resolver.js
```diff
@@ -25,10 +25,16 @@ exports.resolve = (binaryBlob, path) => {
   while (parts.length) {
     const key = parts.shift()
     if (node[key] === undefined) {
-      throw new Error(`Object has no property '${key}'`)
+      // Check the named links
+      if (node._namedLinks[key] === undefined) {
```
achingbrain: I wonder if, instead of adding a `_namedLinks` property, you could just memoize the link name -> CID mapping in this function if, for example, `parts.length` was longer than 1.

IPFS rarely traverses `DAGNode`s by path due to HAMTs meaning there's no 1-to-1 mapping so I'm not sure we're gaining much from this.

vmx: `_namedLinks` is also used for `tree()`. I could of course there also traverse the links and extract the name dynamically. But that sounds simpler.

vmx:
There are no named links anymore (you can resolve named links though) since the merge of https://github.com/ipld/js-ipld-dag-pb/pull/151.

PR #141, opened 2019-05-22
Api changes

vmx:
@achingbrain I figured out that doing one PR after another is not a good mode for those changes.

This is all changes I had in mind when I was talking about simplifying the dag-pb API. They can be done as individual PRs as well, in case we want to discuss them individually. This should be seen as a way to discuss if you think we should introduce all those changes or not.

This PR is best reviewed by looking at one commit at a time, they are self-contained. So you might agree on some and disagree on others. I can then PR the ones we agree on.

There's just one thing I go back and forth with is the removal of `toJSON()`. That is from the days where we thought we would have a `toJSON()` call for all IPLD Formats (which we don't have). I can see that it is useful for our tests (it's used there heavily to compare results), but I think we shouldn't have a public API just for testing purpose (on the other hand, `toJSON()` also doesn't to much harm either).


README.md
```diff
@@ -139,68 +133,69 @@ links can be a single or an array of DAGLinks instances or objects with the foll
...
-```JavaScript
-const dagNode = await DAGNode.rmLink(node, 'Link1')
-```
+#### `node.toString()`
 
-#### clone(node)
+#### `node.toDAGLink()`
 
-- `node` - type: DAGNode
+- `options` - (optionsal) type: Object. Currently the only option is `name` to specify a named link.
```
achingbrain: Typo: optionsal

achingbrain:
Could named links be removed?  I'm importing large datasets and the code spends almost 8% of it's time looping over the links and it's not a feature that's used anywhere:

image

vmx:
@achingbrain I'm open to remove the named links, it should really be on another layer and not within the dag-pb codec. I'm checking with the Go IPLD team though.

What's the plan with this PR. should we go with those changes? Or parts of them? I'm not a fan of long living, huge, rotting PRs.

vmx:
@achingbrain Here's a new version of all this. I think it's best reviewed commit by commit. These are all changes I have in mind to have a dag-pb which won't change much in the future.

src/dag-node/toDagLink.js
```diff
@@ -1,14 +1,15 @@
...
+const genCid = require('../genCid')
 
 /*
  * toDAGLink converts a DAGNode to a DAGLink
  */
 const toDAGLink = async (node, options = {}) => {
-  const serialized = serialize(node)
-  const nodeCid = await cid(serialized)
+  const serialized = serializeDAGNode(node)
+  const nodeCid = await genCid.cid(serialized)
```
achingbrain: It'd be great if we could avoid the double-serialization thing.

The typical usecase is having just `ipld.put`'d a node, you then want to add it as a link to a parent node, for which you need the `cid` and the serialized size.

The `cid` is returned by `ipld.put` but the serialized size is not available.  I suppose you could ask the block store directly, but it'd need some sort of lru cache so it wouldn't go to disk for every request.

Better yet would be to remove the idea of sizes from `DAGLink`s as they are unverifiable anyway.

vmx: > Better yet would be to remove the idea of sizes from `DAGLink`s as they are unverifiable anyway.

Is the `Tsize` in the `PBLink` the same as the `DAGLink` size? If yes, would it mean that you would need to change UnixFS in order to get rid of the `DAGLink` size?

Can this problem be discussed outside this PR? I'd really like to move the dag-pb API changes forward.

achingbrain: > Would it mean that you would need to change UnixFS in order to get rid of the DAGLink size?

Yes, I believe so though I think it's no bad thing - we don't use that field because the useful thing is the size of the file, not the size of the DAG that holds the file.

> Can this problem be discussed outside this PR?

Absolutely, I didn't mean to suggest we should resolve that here.

src/dag-node/addLink.js
```diff
@@ -1,39 +1,33 @@
...
   }
 
-  if (DAGNode.isDAGNode(link)) {
-    // It's a DAGNode instance
-    // convert to link
-    return toDAGLink(link, {})
+  // DAGNode.isDagNode() would be more appropriate here, but it can't be used
+  // as it would lead to circular dependencies as `addLink` is called from
+  // within the DAGNode object.
+  if (!('cid' in link ||
```
rvagg: How much does size matter in a DAGLink? The name property is optional (`|| ''`) but the size property isn't, it just uses what it's given as the second argument. So below, `link.Tsize || link.size` seems like it's assuming something useful is there? Should it be tested for?

`!('cid' in link || 'hash' in link || 'Hash' in link || 'multihash' in link) && !('Tsize' in link || 'size' in link)`

Or maybe even stricter, `!(Number.isInteger(link.Tsize) || Number.isInteger(link.size))`?

vmx: Good point! Given that we want (and hopefully will) get rid of the `size` I dare to leave it like this for now and merge this PR as it is.

PR #146, opened 2019-05-27
chore: remove dependency on async

vmx:
As we are moving from callbacks to async/await, it makes sense to
remove all occurrences of the async module.


achingbrain:
FWIW `ipfs-repo@0.27.0` is now all `async`/`await` so no need to promisify everything any more.

vmx:
@achingbrain Ready for review!

PR #149, opened 2019-06-25
chore: refactor internal circular dependencies

cdata:
## Description

This change proposes the nominal refactoring of the `dag-link` and `dag-node` internal modules to remove the following circular dependencies:

 - `src/dag-link/index.js => src/dag-link/util.js => src/dag-link/index.js`
 - `src/dag-node/index.js => src/dag-node/create.js => src/util.js => src/dag-node/index.js`
 - `src/dag-node/index.js => src/dag-node/addLink.js => src/dag-node/index.js`

## Rationale

Some JavaScript bundlers do not guarantee graceful handling of circular dependencies. One such example is the popular bundler [Rollup](https://github.com/rollup/rollup) ([example issue](https://github.com/rollup/rollup/issues/1507)).

In many cases, fixing circular dependencies is just a matter of lightly refactoring common dependencies into their own modules (which is the nature of this change).

With this change in place, more bundlers should be able to resolve references to all dependencies in the correct order.


src/dag-node/addLink.js
```diff
@@ -5,7 +5,7 @@ const cloneLinks = dagNodeUtil.cloneLinks
 const cloneData = dagNodeUtil.cloneData
 const toDAGLink = dagNodeUtil.toDAGLink
 const DAGLink = require('../dag-link')
-const DAGNode = require('./index')
+const DAGNode = require('./dagNode.js')
```
vmx: We normally don't require with the `.js` extension. I know that in this code base we are not consistent, but it would be great to do this for new changes.

So please change this, if you're making changes anyway. If you don't, nevermind it's not a blocker for getting merged.

src/dag-node/create.js
```diff
@@ -6,7 +6,7 @@ const {
 } = require('../util.js')
 const dagNodeUtil = require('./util.js')
 const linkSort = dagNodeUtil.linkSort
-const DAGNode = require('./index.js')
+const DAGNode = require('./dagNode.js')
```
vmx: Same here, remove the `.js`. But it's not a blocker for being merged.

src/util.js
```diff
@@ -4,7 +4,7 @@ const CID = require('cids')
 const protons = require('protons')
 const proto = protons(require('./dag.proto.js'))
 const DAGLink = require('./dag-link')
-const DAGNode = require('./dag-node')
+const DAGNode = require('./dag-node/dagNode.js')
```
vmx: Same as above, please remove the `.js`.

In regards to the line above, would it make sense to change it to:

```JS
   const DAGLink = require('./dag-link/dagLink')
```

for more consistency?

cdata:
Thanks for the feedback @vmx . I removed all of the `.js` extensions that I could find from module specifiers, and made the specifiers for `DAGLink` more explicit. PTAL at your leisure!

PR #151, opened 2019-06-26
perf: remove named links from object

vmx:
This change was motivated by performance issues, but it also
really is cleaner to keep the IPLD objects lean and not cluttered
with IPFS related things.

BREAKING CHANGE: named links are no longer part of an object

Access to named links is only possible with calling `resolve()`.
Hence they are also not part of `tree()` anymore.

Named links are a feature of IPFS and only supported for
backwards compatibility, they are not really part of IPLD.


mikeal:
> Named links are a feature of IPFS and only supported for backwards compatibility, they are not really part of IPLD.

It’s just semantics, but I’d say they are part of `dag-pb` rather than IPFS. As I continue to discover, people are doing stuff with `dag-pb` by itself in a bunch of places and they may be using this feature, which means they’re going to have a rough upgrade path to IPLD Data Model if they ever decide to go that route.

vmx:
> It’s just semantics, but I’d say they are part of `dag-pb` rather than IPFS. As I continue to discover, people are doing stuff with `dag-pb` by itself in a bunch of places and they may be using this feature

Basically this gets us to how version 0.15.0 and earlier worked. Having named links on the object was a recent addition.

> which means they’re going to have a rough upgrade path to IPLD Data Model if they ever decide to go that route.

I don't get that part. Are you saying they have a rough upgrade if they use named links or if this PR is merged?

mikeal:
> I don't get that part. Are you saying they have a rough upgrade if they use named links or if this PR is merged?

I should have been clearer, I meant “if they used named links.” This PR shouldn’t effect them.

PR #156, opened 2019-08-09
fix: serialization portability problems and size calculation bug

rvagg:
2 bugs I discovered while trying to write my zip archive format for non-trivial portable test fixture data (i.e. my JS version of https://github.com/rvagg/go-ds-zipcar wasn't giving me the same CIDs).

1. `DAGNode` used to be immutable, `addLink()` and `rmLink()` would return new nodes on their callbacks. The current incarnation, since #141, mutates the current node, this means that `node._serializedSize` is incorrect upon either mutation operation. To fix this I've reset the cached sizing data and it's lazily re-calculated when you call `node.size` (as it is if you call the constructor without `serializedSize`).
2. `DAGNode#toDAGLink()` uses _serialized node size_ rather than _node size_ to create the new link, this is not how the Go implementation does it. That happened in #127 and I think it was simply a mistake.

Since `node._serializedSize` needs to be calculated lazily now (sometimes it's never set, sometimes the node mutates), I added a `node.serialize()` to do the work which makes for a really nice API instead of having to go through `dagPB.util.serialize(node)` (I wouldn't mind a `DAGNode.deserialize(buf)` too to get rid of the need to use `dagPB.util` entirely tbh but maybe that can come later).

Also added some Go code to the test/ dir that generates the fixture data I used in the tests I added.

I'd normally just call the above two items critical bug fixes but I've labelled them "BREAKING CHANGE" because I'm not sure what the criteria for "breaking" is in this repo.

-------

    feat: DAGNode#serialize() is now a synonym for
    dagPB.util.serialize(node).

    BREAKING CHANGE: `node.addLink()` and `node.rmLink()` (node mutation
    operations) will now reset any cached size data causing sizing and
    serialization to be performed from the new state when required.

    BREAKING CHANGE: (fix) `node.toDAGLink()` (and therefore
    `node.addLink(DAGNode)` indirectly) now use the `node.size()` value for
    link size rather than the node's serialized byte size, yielding different
    serialized nodes and CIDs. This matches the Go implementation, see
    github.com/ipfs/go-ipld-format's `node.MakeLink()`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,8 +133,6 @@ links can be a single or an array of DAGLinks instances or objects with the foll
 }
 ```
 
-
-
 ### DAGNode instance methods and properties
 
 You have the following methods and properties available in every DAGNode instance.
@@ -166,7 +164,7 @@ const link = node.toDAGLink()
 const link = node.toDAGLink({ name: 'name-of-the-link' })
 ```
 
-#### addLink(link)
+#### `node.addLink(link)`
 
 - `link` - type: DAGLink or DAGLink in its json format
 
@@ -187,7 +185,7 @@ const link = {
 node.addLink(link)
 ```
 
-#### rmLink(nameOrCid)
+#### `node.rmLink(nameOrCid)`
 
 - `nameOrCid` - type: String, CID object or CID buffer
 
@@ -197,6 +195,9 @@ Removes a link from the node by name. Modifies the node.
 node.rmLink('Link1')
 ```
 
+#### `node.serialize()`
+
+Serialize the DAGNode instance to its portable binary format. Yields the same result as `dagPB.util.serialize(node)`. Returns a `Buffer`.
 
 ### DAGLink functions
 
@@ -249,7 +250,11 @@ const link = new DAGLink(
 
 ### `dagPB.util.serialize`
 
+Serialize the DAGNode instance to its portable binary format. Yields the same result as `node.serialize()`. Returns a `Buffer`.
+
 ### `dagPB.util.deserialize`
+
+Deserialize a DAGNode instance from its portable binary format. Returns a DAGNode.
 
 ## Contribute
 
--- a/src/dag-node/dagNode.js
+++ b/src/dag-node/dagNode.js
@@ -9,7 +9,7 @@ const addLink = require('./addLink')
 const rmLink = require('./rmLink')
 
 class DAGNode {
-  constructor (data, links = [], serializedSize = 0) {
+  constructor (data, links = [], serializedSize = null) {
     if (!data) {
       data = Buffer.alloc(0)
     }
@@ -20,6 +20,10 @@ class DAGNode {
       throw new Error('Passed \'data\' is not a buffer or a string!')
     }
 
+    if (serializedSize !== null && typeof serializedSize !== 'number') {
+      throw new Error('Passed \'serializedSize\' must be a number!')
+    }
+
     links = links.map((link) => {
       return DAGLink.isDAGLink(link)
         ? link
@@ -27,16 +31,10 @@ class DAGNode {
     })
     links = sortLinks(links)
 
-    if (serializedSize === 0) {
-      serializedSize = serializeDAGNode({
-        Data: data,
-        Links: links
-      }).length
-    }
-
     this._data = data
     this._links = links
     this._serializedSize = serializedSize
+    this._size = null
   }
 
   toJSON () {
@@ -55,11 +53,18 @@ class DAGNode {
     return `DAGNode <data: "${this.Data.toString('base64')}", links: ${this.Links.length}, size: ${this.size}>`
   }
 
+  _invalidateCached () {
+    this._serializedSize = null
+    this._size = null
+  }
+
   addLink (link) {
+    this._invalidateCached()
     return addLink(this, link)
   }
 
   rmLink (link) {
+    this._invalidateCached()
     return rmLink(this, link)
   }
 
@@ -68,8 +73,18 @@ class DAGNode {
     return toDAGLink(this, options)
   }
 
+  serialize () {
+    return serializeDAGNode({
+      Data: this._data,
+      Links: this._links
+    })
+  }
+
   get size () {
-    if (this._size === undefined) {
+    if (this._size === null) {
+      if (this._serializedSize === null) {
+        this._serializedSize = this.serialize().length
+      }
       this._size = this._links.reduce((sum, l) => sum + l.Tsize, this._serializedSize)
     }
 
--- a/src/dag-node/toDagLink.js
+++ b/src/dag-node/toDagLink.js
@@ -1,16 +1,14 @@
 'use strict'
 
 const DAGLink = require('../dag-link/dagLink')
-const { serializeDAGNode } = require('../serialize')
 const genCid = require('../genCid')
 
 /*
  * toDAGLink converts a DAGNode to a DAGLink
  */
 const toDAGLink = async (node, options = {}) => {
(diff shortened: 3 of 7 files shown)


src/dag-node/toDagLink.js
```diff
@@ -1,16 +1,14 @@
...
 const genCid = require('../genCid')
 
 /*
  * toDAGLink converts a DAGNode to a DAGLink
  */
 const toDAGLink = async (node, options = {}) => {
-  const serialized = serializeDAGNode(node)
-  const nodeCid = await genCid.cid(serialized)
-  return new DAGLink(options.name || '', serialized.length, nodeCid)
+  const nodeCid = await genCid.cid(node.serialize(), options)
```
rvagg: oh yeah, passing `options` here is a new feature, is it worth mentioning? basically I wanted to be able to get v0 CIDs out of a `addLink(node)` instead of going indirectly through `util.cid()`.

src/dag-node/dagNode.js
```diff
@@ -27,16 +27,10 @@ class DAGNode {
...
-      serializedSize = serializeDAGNode({
-        Data: data,
-        Links: links
-      }).length
-    }
-
     this._data = data
     this._links = links
-    this._serializedSize = serializedSize
+    this._serializedSize = typeof serializedSize === 'number' ? serializedSize : null
```
vmx: `serializedSize` isn't manipulated within the constructor. Would it work to change its signature to `constructor (data, links = [], serializedSize = null)` and then do a `this._serializedSize = serializedSize` here?

test/pbcross.go
```diff
@@ -0,0 +1,46 @@
+// `go get github.com/ipfs/go-merkledag`
```
vmx: I'm not sure if this should be in the `test` directory. Perhaps we create a new directory called `tools` ([js-multicodec also has a tools directory](https://github.com/multiformats/js-multicodec/tree/86a6b84ce176fed29b2e1d3f7698e40f6fb8dc79/tools) (or any better name you come up with) to store this file, so people don't wonder why there's a Go source file in their test directory.

Or if those fixtures are already somewhere else (e.g. in go-car), we could just put a link to it in the js test and interested people can go from there.

README.md
```diff
@@ -197,6 +195,9 @@ Removes a link from the node by name. Modifies the node.
 node.rmLink('Link1')
 ```
 
+#### `node.serialize()`
```
vmx: I was skeptical about having a `node.serialize()` as I'm not a fan of having two ways of doing the same thing. Though I needed to do weird serializer import tricks due to circular dependencies which might not be needed anymore. Hence I'm supportive of this change. Therefore I think it also makes sense to add `node.deserialize()` for API symmetry in a subsequent PR.

Also `node.serialize()` can be seen as the DAP-PB API and the `util.*` calls are just the compatibility layer to make it work with js-ipld.

src/dag-node/dagNode.js
```diff
@@ -68,9 +69,19 @@ class DAGNode {
...
+  }
+
   get size () {
-    if (this._size === undefined) {
-      this._size = this._links.reduce((sum, l) => sum + l.Tsize, this._serializedSize)
+    if (this._size === null) {
+      if (this._serializedSize === null) {
+        this._serializedSize = this.serialize().length
+      }
+      this._size = this._links.reduce((sum, l) => sum + l.Tsize, this.serialize().length)
```
vmx: `this.serialize().length` is called whenever `this._size === null`. This means that we could just unconditionally assign it to `this._serializedSize`:

```js
get size() {
  if (this._size === null) {
    this._serializedSize = this.serialize().length
    this._size = this._links.reduce((sum, l) => sum + l.Tsize, this._serializedSize)
  }
}
```

src/util.js
```diff
@@ -54,7 +54,7 @@ const deserialize = (buffer) => {
 
   const data = pbn.Data == null ? Buffer.alloc(0) : pbn.Data
 
-  return new DAGNode(data, links, buffer.length)
+  return new DAGNode(data, links, buffer)
```
vmx: This should be `DAGNode(data, links)` if the constructor is modified as suggested in my other review comment.

src/dag-node/dagNode.js
rvagg: I did this because you have no guarantees of type even with a default argument value, in this way, it's either `null` or a number, no options for anything else. It's not a big deal, I just prefer more assurance of type and default values don't give you much of that. Your call.

src/util.js
rvagg: good catch, comes from an intermediate version where I cached the entire `buffer` instead of just size but then I found git history where it was like that once but changed for memory reasons

test/pbcross.go
rvagg: moved to tools. go-car doesn't separate them out like this, I figure it's nice to have a proper reference so if we come across something new that's broken between the two implementations we can add it to the .go implementation

src/dag-node/dagNode.js
vmx: I see. It find it strange if you have a default value as the second parameter, but no default value for the third. So I would prefer having at least a `serialized = null` in the constructor.

When reading the code I didn't get that it was there as a sanity check. We have some checks at the top, so I would rather put a check like

```js
if (serializedSize !== null && serializedSize !== number) {
   throw new Error('Passed \'serializedSize\' must be a number')
}
```

No sure about this though, so take this as input and I you "making the call" back to you.

src/dag-node/dagNode.js
rvagg: oh, that's a mistake, changed it to have only that one call to `this.serialize().length`:

```
    if (this._size === null) {
      if (this._serializedSize === null) {
        this._serializedSize = this.serialize().length
      }
      this._size = this._links.reduce((sum, l) => sum + l.Tsize, this._serializedSize)
    }
```

If `this._serializedSize` comes in from the constructor then we only do the outer condition, if it's not in the constructor then we do both. The use in `util.deserialize()` to send `buffer.length` (unlike my mistaken `buffer`, see below) in is still a good optimisation I think and saves a re-serialisation for this size calculation.

Mind you, I have no idea how much this matters, how often is `size()` calculated on real data? How much does it really matter? I just don't have insight to how this is used upstream.

rvagg:
fixed those two glaring mistakes (`buffer.length` vs `buffer` and a double call to `this.serialize()`) in feb36aff

src/dag-node/dagNode.js
vmx: I also have no idea. It just seemed easy enough to have it called only once.

rvagg:
sorry, it's a24a8d2, might want to review that commit as a stand-alone as there's a change in the test too.

src/dag-node/dagNode.js
rvagg: I like JS APIs to be very explicit about types so I'm +1 on this, it's just a bit more "breaking" than allowing a silent ignore. But I suppose that third argument isn't intended for public use so it's safe to be brutal about it?

rvagg:
OK, ready again, I went with the strict `!=null&&!'number'` test in the constructor

vmx:
Would you mind force-pushing a fixed-up commit?

vmx:
I forgot to get into your question about breaking changes. I won't consider them a "BREAKING CHANGE" as I use those to signal that the user using this library might need to make changes to their code if they upgrade to the newer version.

In this case I'd expect to have information about those fixes in the commit message, but not marked as "BREAKING CHANGE". I'd then do a bugfix release.

rvagg:
autosquashed, edited the commit message and force pushed

PR #164, opened 2020-01-15
feat: ignore tsize property on pblinks

### LINKED ISSUE #154: Prevent double serialization
This is a followup of https://github.com/ipld/js-ipld-dag-pb/pull/141#pullrequestreview-263570966.

`toDagLink()` should double serialize things. Best would be if `size` would be removed from DAGLink. Though this is a change that impacts UnixFS.

achingbrain:
The size property has been removed from links. In order to maintain backwards compatibility the field is stil present in the protobuf definition but is ignored.

See discussion on https://github.com/ipfs/specs/issues/238 and https://github.com/ipld/specs/pull/234

Fixes #154

BREAKING CHANGE:

* The DAGLink constructor has changed from `DAGLink(name, size, cid)` to `DAGLink(name, cid)`
* The DAGNode constructor has changed from `DAGNode(data, links, serializedSize)` to `DAGNode(data, links)`
* The DAGLink class no long has a `.Tsize` property
* The DAGNode class no longer has `.size` or `._serializedSize` properties

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# js-ipld-dag-pb
+# js-ipld-dag-pb <!-- omit in toc -->
 
 [![](https://img.shields.io/badge/made%20by-Protocol%20Labs-blue.svg?style=flat-square)](http://ipn.io)
 [![](https://img.shields.io/badge/project-IPLD-blue.svg?style=flat-square)](http://github.com/ipld/ipld)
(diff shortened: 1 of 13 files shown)


alanshaw:
Suggestion: `DAGLink(cid, name)` over `DAGLink(name, cid)`?

rvagg:
I guess it depends on the outcome of https://github.com/ipld/specs/pull/234, other than that I'll defer to your collective expertise on this one.

But while we're doing this, can we please bump major so we can properly signal these kinds of breaking changes into the future.

achingbrain:
> can we please bump major so we can properly signal these kinds of breaking changes

Yes please!  See: https://github.com/ipfs/js-ipfs/issues/2657

achingbrain:
The outcome is we're still going to require generating the `Tsize` so this PR should not be merged.

PR #167, opened 2020-02-14
fix: remove use of assert module

achingbrain:
The polyfill is big, we can simulate it by throwing an Error and it doesn't work under React Native.


achingbrain:
Node, browsers, React Native, Electron.  It's all reminding me of ['Write once, debug everywhere'](https://en.wikipedia.org/wiki/Write_once,_run_anywhere).

Issue #173, opened 2020-03-20
Migrate to a pure Data Model API


mikeal:
This would be a breaking change.

I’d like to use the new [IPLD Schema for `dag-pb`](https://github.com/ipld/specs/pull/247) for this codec.

This would be a major API break, but we probably need to do it because it would:

* make it much easier to use the codec since you’d be able to interact with regular JS objects rather `Node` instances
* align with the other implementations of `dag-pb`
* allow for selector engine support once the JS selector engine lands

However, this would be a pretty big break for `js-ipfs`. We’d want to time this along with a larger migration to the `Block` API.

/cc @achingbrain @vmx @alanshaw @carsonfarmer

carsonfarmer:
Aside from the breaking changes to IPFS, I would say "yes please" to this proposal! Already most consumers of this module need to handle the 'Node' special-case to do the right thing, so it would be really nice to shed some of that cognitive overhead. We're already using the Block API in our internal (Textile) work, so less of a 'breaking' issue and more of a 'fixing' thing for us TBH.

Gozala:
@mikeal can you clarify what's backwards incompatible about this change ? Talking to @rvagg I'm under impression that that this can be in fact backwards compatible change.

As I understand it it's merely treating anything shaped as

```ts
{
  Data?: Uint8Array,
  Links?: Array<{ Hash: Uint8Array, Name?:string, Tsize?: number }>
}
```

If that is indeed the case, most of the utils already work under than assumption. So it would be just matter of making corresponding getters an actual fields. That way `DAGNode` can still be kept around for backwards compatibility but can effectively be ignored.

If that is indeed this is all about, I'm happy to take it as that would unblock my work described in https://github.com/ipld/js-ipld/issues/283

Gozala:
As a side note I've just noticed that this implementation does seem to expect `Buffer` and fails on `Uint8Array` (because it's assumes it's a string). I would like to address that as well because once `Buffer`s move across threads they turn into Uint8Array`s and I'd rather make dag-pb capable of handling `Uint8Array`s than doing conversations the thread messaging layer.

rvagg:
+1, all of our newest code is trying to be `Buffer` free where possible. js-multiformats doesn't use it all.

rvagg:
[...]

PR #184, opened 2020-06-05
Backwards compatible pure data model API

Gozala:
This pull request attempts to add pure Data Model API interop (as per #173) in a backwards compatible way. High level overview of the changes are:

- Getters / Setters are replaced by non-writable public properties.
- Some internal ops that add / remove links do mutate arrays in place instead of swapping arrays (due to introduce non-writable constraint).
- serialization code now handles ArrayBuffer and TypedArray for `node.Data`.

Only visible change is that `node.Links` no longer returns links mapped to pure data, but rather `DAGLink` instances. However it should not matter in practice because those were made to behave just like pure data would. Although tests had to be change to use `.containSubset` instead of `.eql` because later fails when prototype chain is different.

### DIFF
--- a/package.json
+++ b/package.json
@@ -76,6 +76,7 @@
     "aegir": "^22.0.0",
     "chai": "^4.2.0",
     "dirty-chai": "^2.0.1",
+    "chai-subset": "1.6.0",
     "fs-extra": "^9.0.0",
     "ipfs-block-service": "~0.17.0",
     "ipfs-repo": "^3.0.0",
--- a/package.json
+++ b/package.json
@@ -74,9 +74,6 @@
   },
   "devDependencies": {
     "aegir": "^22.0.0",
-    "chai": "^4.2.0",
-    "dirty-chai": "^2.0.1",
-    "chai-subset": "1.6.0",
     "fs-extra": "^9.0.0",
     "ipfs-block-service": "~0.17.0",
     "ipfs-repo": "^3.0.0",
--- a/package.json
+++ b/package.json
@@ -69,7 +69,7 @@
     "class-is": "^1.1.0",
     "multicodec": "^1.0.1",
     "multihashing-async": "~0.8.1",
-    "protons": "^1.0.2",
+    "protons": "git://github.com/gozala/protons#uint8array",
     "stable": "^0.1.8"
   },
   "devDependencies": {
--- a/package.json
+++ b/package.json
@@ -69,7 +69,7 @@
     "class-is": "^1.1.0",
     "multicodec": "^1.0.1",
     "multihashing-async": "~0.8.1",
-    "protons": "git://github.com/gozala/protons#uint8array",
+    "protons": "^1.2.1",
     "stable": "^0.1.8"
   },
   "devDependencies": {
--- a/src/dag-node/dagNode.js
+++ b/src/dag-node/dagNode.js
@@ -25,24 +25,26 @@ class DAGNode {
       throw new Error('Passed \'serializedSize\' must be a number!')
     }
 
-    links = links.map((link) => {
+    links = links.map(link => {
       return DAGLink.isDAGLink(link)
         ? link
         : DAGLink.util.createDagLinkFromB58EncodedHash(link)
     })
-    links = sortLinks(links)
+    sortLinks.inplace(links)
 
-    this._data = data
-    this._links = links
-    this._serializedSize = serializedSize
-    this._size = null
+    Object.defineProperties(this, {
+      Data: { value: data, writable: false, enumerable: true },
+      Links: { value: links, writable: false, enumerable: true },
+      _serializedSize: { value: serializedSize, writable: true, enumerable: false },
+      _size: { value: null, writable: true, enumerable: false }
+    })
   }
 
   toJSON () {
     if (!this._json) {
       this._json = Object.freeze({
         data: this.Data,
-        links: this._links.map((l) => l.toJSON()),
+        links: this.Links.map((l) => l.toJSON()),
         size: this.size
       })
     }
@@ -75,10 +77,7 @@ class DAGNode {
   }
 
   serialize () {
-    return serializeDAGNode({
-      Data: this._data,
-      Links: this._links
-    })
+    return serializeDAGNode(this)
   }
 
   get size () {
@@ -86,7 +85,7 @@ class DAGNode {
       if (this._serializedSize === null) {
         this._serializedSize = this.serialize().length
       }
-      this._size = this._links.reduce((sum, l) => sum + l.Tsize, this._serializedSize)
+      this._size = this.Links.reduce((sum, l) => sum + l.Tsize, this._serializedSize)
     }
 
     return this._size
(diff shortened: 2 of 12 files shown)


vmx:
Wow, nice work. I just tested it also with `js-ipld` and it just worked. If @achingbrain is happy with those changes, I don't see a reason for not making those changes.

test/dag-node-test.js
```diff
@@ -4,8 +4,10 @@
 const chai = require('chai')
 const { Buffer } = require('buffer')
 const dirtyChai = require('dirty-chai')
+const chaiSubset = require('chai-subset')
```
achingbrain, line 7: You can require chai from `aegir/utils/chai` and it has `dirty-chai` and `chai-subset` preconfigured. It also avoids nasty weird bugs related to the order of `chai.use` invocations [across the entire codebase](https://github.com/chaijs/chai/issues/1298).

test/dag-node-test.js
```diff
@@ -80,7 +82,7 @@ module.exports = (repo) => {
       })
 
       const node2 = new DAGNode(someData, l2)
-      expect(node2.Links).to.eql([l1[1], l1[0]])
+      expect(node2.Links).to.containSubset([l1[1], l1[0]])
```
achingbrain, line 85: Doesn't marking the properties non-enumerable mean you should be able to use `.deep.equal` (or `.eql`) here?

src/serialize.js
```diff
@@ -20,7 +20,7 @@ const toProtoBuf = (node) => {
   if (node.Links && node.Links.length > 0) {
     pbn.Links = node.Links
       .map((link) => ({
-        Hash: link.Hash.buffer,
+        Hash: asBuffer(link.Hash.buffer),
```
achingbrain, line 23: Did you mean to pass `link.Hash` in here instead?

If `link.Hash` is a node buffer, the `.buffer` property is an ArrayBuffer that might not correspond exactly to the original bytes.

We need to use `Buffer.from`, `link.Hash.byteOffset` and `link.Hash.length` to make sure we don't end up with the wrong bytes if we're accessing the backing ArrayBuffer directly - https://nodejs.org/api/buffer.html#buffer_buf_buffer

As it stands it'll get turned back into a node Buffer further down at line 44 that contains the whole ArrayBuffer.

src/dag-node/sortLinks.js
```diff
@@ -8,12 +8,23 @@ const linkSort = (a, b) => {
 }
 
 /**
- *
+ * Returns new sorted links array.
  * @param {Array} links
  * @returns {Array}
  */
 const sortLinks = (links) => {
```
achingbrain, line 15: If we're switching to in-place sorting, is this function used somewhere else in the module or can it be removed?

test/dag-node-test.js
Gozala: I believe problem with `.deep.equal` and `.eql` is that fails if values do have different prototype chains.

src/serialize.js
Gozala: Good catch, I meant `link.Hash` indeed (I think I've overlooked `link.Hash` was Buffer rather than thing that had `{buffer: Buffer}`).

_Fact that no test caught this makes me quite a bit nervous._

src/dag-node/sortLinks.js
Gozala: Only reason I have not done it is I do not know if anything outside of the package might be making use of this. If we're comfortable removing it, I'm happy to do so

src/serialize.js
Gozala: Wait no, `link.Hash` is a `CID` not the `Buffer`

Gozala: New commit removes this changes in favor of adding Uint8Array support into protons library instead (See https://github.com/ipfs/protons/pull/13)

test/dag-node-test.js
Gozala: Turns out the problem here was caused by `[Symbol.toStringTag]` getter added by `class-is` library.

https://github.com/moxystudio/js-class-is/blob/master/index.js#L19-L21

I'm assuming `eql` somewhere does `Object.prototype.toString.call(v)` but unless this is a blocker, I would prefer not to dig into this any further.

src/dag-node/sortLinks.js
Gozala: Made a change that makes replaces existing `sortLinks` with mutable sortLinks`. One thing that makes me bit uncomfortable about this change is that it's default import and change will in behavior will not be immediately obvious (if it was named, I'd just renamed the function).

Gozala:
Summary of few updates I've made:

1. Pull `chai` from `aegir`.
2. Remove `Uint8Array` to `Buffer` casting in favor of doing it in protons (See https://github.com/ipfs/protons/pull/13)
  - Current patch changes protons dep to point to my fork. Once it lands and published I'll update this to point to new verison.
  - `ArrayBuffer` as `Data` is no longer supported, it's just `Uint8Array` now. Mostly to keep things simple on `protons` side.
3. Replaced immutable `sortLinks` with mutable version. Immutable version was not used anywhere else in this package.

package.json
```diff
@@ -69,13 +69,11 @@
     "class-is": "^1.1.0",
     "multicodec": "^1.0.1",
     "multihashing-async": "~0.8.1",
-    "protons": "^1.0.2",
+    "protons": "git://github.com/gozala/protons#uint8array",
```
vmx, line 72: I planned to switch to https://github.com/mapbox/pbf if we ever switch to UInt8Array support. It might be worth a look now. IIRC that version just works automatically on any array-like thing, hence Node.js Buffers as well as UInt8Arrays.

test/dag-node-test.js
achingbrain: Could you open an issue against `class-is` please?  It should not get in the way like this.

src/serialize.js
achingbrain: > Wait no, link.Hash is a CID not the Buffer

Quite right, this is why I shouldn't look at these things in the evening!

package.json
Gozala: That sounds like a good idea! However I would much rather keep the rabbit hole as shallow as possible. That is to say can we do that independently of this pull request ?

P.S.: I obviously don't suggest landing this with my fork in dependencies, just have it here until this https://github.com/dignifiedquire/borc/pull/48 makes it's way to npm.

src/dag-node/rmLink.js
```diff
@@ -4,12 +4,26 @@ const CID = require('cids')
 const { Buffer } = require('buffer')
 
 const rmLink = (dagNode, nameOrCid) => {
+  let predicate = null
+
   // It's a name
   if (typeof nameOrCid === 'string') {
-    dagNode._links = dagNode._links.filter((link) => link.Name !== nameOrCid)
+    predicate = link => link.Name === nameOrCid
```
rvagg, line 11: minor nit: the style in this library is to always-paren arrow functions, it'd be great to keep it that way (a rule I'd love enforced by standard tbh, for readability, but for now it's on a per-library convention basis). There's another instance 2 lines below too.

src/dag-node/dagNode.js
```diff
@@ -25,24 +25,26 @@ class DAGNode {
       throw new Error('Passed \'serializedSize\' must be a number!')
     }
 
-    links = links.map((link) => {
+    links = links.map(link => {
```
rvagg, line 28: another instance of missing parens

test/dag-node-test.js
Gozala: https://github.com/moxystudio/js-class-is/issues/27

Gozala:
@achingbrain I believe I've addressed all your comments and comments from @rvagg.  @vmx I hope you agree with my arguments to decouple this from switch to https://github.com/mapbox/pbf, I do not know what errors may arise from that change & I do not want to block my primary work on that.

Please let me know if there is anything else to be done, or if this can land.

Thanks

Gozala:
@vmx any chance you could get to this sometime this week ? I can't make any more progress on  ipfs/js-ipfs#3081 without this.

vmx:
@Gozala any objections if I release this as 0.19? It's been quite a big change and I don't want to have regressions sneaked it.

Gozala:
@vmx that sounds good to me. Thanks

PR #188, opened 2020-11-23
fix: strip control characters from link names

achingbrain:
In the case where we receive garbage or malicious strings that contain
characters with a utf code less than 31, strip those character codes
from the link name.

We keep the garbage string around though, so we can de-serialize and
reserialize a node without changing it's CID, though we do not expose
that string to the caller through documented properties.

### DIFF
--- a/src/dag-link/dagLink.js
+++ b/src/dag-link/dagLink.js
@@ -16,10 +16,11 @@ class DAGLink {
     //  for now to maintain consistency with go-ipfs pinset
 
     Object.defineProperties(this, {
-      Name: { value: name || '', writable: false, enumerable: true },
+      Name: { value: stripControlChars(name) || '', writable: false, enumerable: true },
       Tsize: { value: size, writable: false, enumerable: true },
       Hash: { value: new CID(cid), writable: false, enumerable: true },
-      _nameBuf: { value: null, writable: true, enumerable: false }
+      _nameBuf: { value: null, writable: true, enumerable: false },
+      _name: { value: name, writable: false, enumerable: false }
     })
   }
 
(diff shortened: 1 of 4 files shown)


vmx (approved):
Makes sense to me. Is that something that other IPFS implementations also plan to do?

warpfork:
Dislike.

When asking for this property, it's not guaranteed that the next thing that's going to happen is printing out to some context where these control characters will actually have an effect.  So enforcing this on all consumers of this API seems incorrect.

We should any kind of escaping or stripping like this as close to the context of _usage_ as possible, not close to the data model or core APIs.

warpfork:
Discussed in the IPLD team call this week, at about 46 minutes in.  (Will edit this will link as soon as it's uploaded.)

vmx (commented):
After talking with the team, I take my approval back. To me the main point is that strings with control characters can probably happen at various places, so the best place to deal with them is before you don't want them to show up (or escape them).

rvagg:
> We should any kind of escaping or stripping like this as close to the context of usage as possible, not close to the data model or core APIs.

The key point for these codecs is that the data model instantiation of them should be able to round-trip back to their original binary form. There should be a two-way relationship between them. In this codec it's a little bit fuzzy because of the history of dag-pb, but in the new dag-pb codec @ https://github.com/ipld/js-dag-pb this rule is as strict as it is for all other codecs - what you get in the data model, instantiated as a JavaScript object, must be what emerges from the data in a way that round-trips and doesn't break data model rules.

So, as Eric said, any transformation should be pushed out as far as possible to the point where it's a problem. Kind of like with error handling where it's best to take that to the edge where it can be appropriately dealt with.

achingbrain:
Fair enough. You wouldn't escape HTML entities at this level either.

PR #190, opened 2021-03-03
chore: update deps

achingbrain:


vmx (changes requested):
I know you don't like commit lints, but I find them useful, so that the changelogs autogen. Can you please add them back in?

achingbrain:
They were removed from aegir.

vmx (approved):
Sad to hear that.

PR #191, opened 2021-03-10
chore: add interface type

achingbrain:
Adds ts type for exported interface

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,7 +67,7 @@
   },
   "dependencies": {
     "cids": "^1.0.0",
-    "interface-ipld-format": "0.0.2",
+    "interface-ipld-format": "^1.0.0",
     "multicodec": "^3.0.1",
     "multihashing-async": "^2.0.0",
     "protobufjs": "^6.10.2",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -67,6 +67,7 @@
   },
   "dependencies": {
     "cids": "^1.0.0",
+    "interface-ipld-format": "0.0.2",
```
rvagg, line 70: where's the source of this? https://github.com/ipld/interface-ipld-format doesn't match https://unpkg.com/browse/interface-ipld-format like the package.json suggests it should.

rvagg (approved):

package.json
achingbrain: You found it, but for others it's here: https://github.com/ipld/interface-ipld-format/pull/56

We should merge that and `v1` it before merging this.

rvagg (approved):
according to spec, `Links` shouldn't be optional, but it's probably going to be too hard to enforce that here, that's what `@ipld/dag-pb` was for, so this is fine

vmx (approved):

PR #193, opened 2021-03-12
fix: types again

achingbrain:
If we require a JS class `Foo`, then export it as a property of an object,
ts thinks we are exporting `typeof Foo` rather than the `Foo` constructor.

At runtime `typeof` gets erased so ts gets upset by things like:

```js
const { DAGNode } = require('ipld-dag-pb')

if (bar instanceof DAGNode) {
  // ...
}
```

..because as far as ts is concerned, `DAGNode` is actually `typeof DAGNode`
and `typeof DAGNode` has been erased since it's a `typeof` so `ipld-dag-pb`
does not export a property called `DAGNode` and kaboom.

The change here is to use the `types.d.ts` to define the `typeof` we are
exporting-but-not-really-exporting-since-it-gets-erased, use a `@typedef`
in the index file to actually export the type for use elsewhere, then rename
the `require`'d DAGNode/DAGLink classes to stop them conflicting with the
`@typedef` tag.

JS is happy, ts is happy.

Phew.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -2,21 +2,22 @@
 
 const resolver = require('./resolver')
 const util = require('./util')
-const DAGNode = require('./dag-node/dagNode')
-const DAGLink = require('./dag-link/dagLink')
+const DAGNodeClass = require('./dag-node/dagNode')
+const DAGLinkClass = require('./dag-link/dagLink')
 
 /**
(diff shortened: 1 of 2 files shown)


vmx (approved):
TypeScript types are really fun. Thank you for figuring all this out.

PR #194, opened 2021-04-06
chore: remove redundant deps

achingbrain:
There seem to be some unused deps in the package.json file so we should remove them.


rvagg (approved):
@vmx can you patch & release please?

vmx:
The CI failure seems real. @achingbrain do you have a look or should I?

achingbrain:
It was real but solved by a transitive dependency update - we're pulling web-encoding out of the stack because the type/browser/main/exports thing is too fragile and we don't need it anyway since TextEncoder/Decoder are global in all environments we support.

vmx (approved):

vmx:
Released!

